PR #10, opened 2015-07-30
added is_subset_opt

llogiq:
This is a modestly optimized `is_subset` method. In my benchmarks, it gets:

```
test is_subset       ... bench:       1,170 ns/iter (+/- 1)
test is_subset_2     ... bench:       1,585 ns/iter (+/- 25)
test is_subset_3     ... bench:       4,863 ns/iter (+/- 8)
test is_subset_4     ... bench:          35 ns/iter (+/- 0)
test is_subset_opt   ... bench:       1,168 ns/iter (+/- 2)
test is_subset_opt_2 ... bench:       1,558 ns/iter (+/- 3)
test is_subset_opt_3 ... bench:       4,824 ns/iter (+/- 7)
test is_subset_opt_4 ... bench:           0 ns/iter (+/- 0)
```

So the wins are fairly small (depending on size), but I think it opens up possible future optimizations. Perhaps you want to replace `is_subset` completely? If so, leave a comment on this PR, and I can push another commit.

(Apart from that, fixed a few warnings)

[Review on Reviewable](https://reviewable.io/reviews/nemo157/roaring-rs/10)

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -36,54 +36,54 @@ fn insert2(b: &mut Bencher) {
 
 #[bench]
 fn is_subset(b: &mut Bencher) {
-	let bitmap: RoaringBitmap<u32> = (1..250).collect();
-	b.iter(|| test::black_box(bitmap.is_subset(&bitmap)))
+    let bitmap: RoaringBitmap<u32> = (1..250).collect();
+    b.iter(|| test::black_box(bitmap.is_subset(&bitmap)))
 }
 
 #[bench]
 fn is_subset_2(b: &mut Bencher) {
-	let sub: RoaringBitmap<u32> = (1000..8196).collect();
-	let sup: RoaringBitmap<u32> = (0..16384).collect();
-	b.iter(|| test::black_box(sub.is_subset(&sup)))
+    let sub: RoaringBitmap<u32> = (1000..8196).collect();
+    let sup: RoaringBitmap<u32> = (0..16384).collect();
+    b.iter(|| test::black_box(sub.is_subset(&sup)))
 }
 
 #[bench]
 fn is_subset_3(b: &mut Bencher) {
-	let sub: RoaringBitmap<u32> = (1000..4096).map(|x| x * 2).collect();
-	let sup: RoaringBitmap<u32> = (0..16384).collect();
-	b.iter(|| test::black_box(sub.is_subset(&sup)))
+    let sub: RoaringBitmap<u32> = (1000..4096).map(|x| x * 2).collect();
+    let sup: RoaringBitmap<u32> = (0..16384).collect();
+    b.iter(|| test::black_box(sub.is_subset(&sup)))
 }
 
 #[bench]
 fn is_subset_4(b: &mut Bencher) {
-	let sub: RoaringBitmap<u32> = (0..17).map(|x| 1 << x).collect();
-	let sup: RoaringBitmap<u32> = (0..65536).collect();
-	b.iter(|| test::black_box(sub.is_subset(&sup)))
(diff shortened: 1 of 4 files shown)


Nemo157:
Cool, I'll have to think about where to put the limit between reduced code and optimised implementations before replacing it.

I've just added an http://editorconfig.org settings file, once you've updated tabs -> spaces I'll merge this.

---

Reviewed 1 of 4 files at r1.
Review status: 1 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[benches/lib.rs, line 88 \[r1\]](https://reviewable.io:443/reviews/nemo157/roaring-rs/10#-JvVmfoMQEAVTGmuEslu)** ([raw file](https://github.com/nemo157/roaring-rs/blob/0ede2f4526e70605e565994c39686ce8242870a0/benches/lib.rs#L88)):</sup>
Could you please update tabs to spaces in these methods.

---

<sup>**[src/imp.rs, line 117 \[r1\]](https://reviewable.io:443/reviews/nemo157/roaring-rs/10#-JvVmbbEl9VNTvza3PE9)** ([raw file](https://github.com/nemo157/roaring-rs/blob/0ede2f4526e70605e565994c39686ce8242870a0/src/imp.rs#L117)):</sup>
Could you please update tabs to spaces in this method.

---

<sup>**[src/lib.rs, line 253 \[r1\]](https://reviewable.io:443/reviews/nemo157/roaring-rs/10#-JvVmhfbVCehwYcgh0By)** ([raw file](https://github.com/nemo157/roaring-rs/blob/0ede2f4526e70605e565994c39686ce8242870a0/src/lib.rs#L253)):</sup>
Could you please update these tabs to spaces.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/nemo157/roaring-rs/10)

Nemo157:
:disappointed: I was expecting reviewable to put those comments on the files.

llogiq:
No big deal, I'll just convert tabs to spaces on all files; expect a commit in about two hours (I'm on the road in the meantime).

llogiq:
Done. Next up, I'd like to benchmark if using iterators could be faster than indices (like you did in `store.rs`' `is_subset(…)`). Perhaps we can shave off another few ns.

Nemo157:
Cool :) I really should get round to adding the overall benchmarks comparing to the Java implementation.

llogiq:
I have an iterator-based solution, but it currently performs slightly worse for 	two of the four benchmarks, and equal on the others. I'll stash it somewhere in case Rust's iterator fusion gets better.

llogiq:
Just for the record, here is the iterator-based code:

```
pub fn is_subset_opt<Size: ExtInt + Halveable>(this: &RB<Size>, other: &RB<Size>) -> bool {
    let (t, o) = (&this.containers, &other.containers);
    if t.len() > o.len() { return false; }
    let (mut ti, mut oi) = (t.iter(), o.iter());
    let (mut tv, mut ov) = (ti.next(), oi.next());
    loop {
        match (tv, ov) {
            (None, _) => { return true; }
            (_, None) => { return false; }
            (Some(ref tc), Some(ref oc)) => {
                if tc.key() < oc.key() { return false; }
                if tc.key() == oc.key() {
                    if !tc.is_subset(oc) { return false; }
                    tv = ti.next();
                }
                ov = oi.next();
            }
        }
    }
}
```

lemire:
@Nemo157  

"overall benchmarks comparing to the Java implementation."

On the Java side, we have done quite a lot of work organizing tests around realistic data sets. Feel free to reuse this work (see jmh directory). We would love to see how well roaring-rs fares compare to the Java version!

PR #32, opened 2016-12-24
Serialization

### LINKED ISSUE #30: Serialize/deserialize from/to the "standard" binary format
We have an "official" Roaring format specification that this library should be able to at least partially support...

https://github.com/RoaringBitmap/RoaringFormatSpec

It is simpler than it sounds.

Interoperability with C, Java and Go might be a nice feature for some users.

Nemo157:
Add standard binary serialization.

Fixes #30


cakey:
Ah, I just realised this makes my hacky changes on my fork redundant :) Thank you!

Issue #38, opened 2017-03-06
u64 support


bvinc:
Am I correct in seeing that u64 bitmaps have been removed?  If so, why?  The commit message said that u64 never worked.  Why didn't it work?

Can this be a feature request for u64 roaring bitmaps?  Or are they gone forever?

Nemo157:
Normal 32-bit Roaring bitmaps split the 32-bit space into 16,384 containers; each containing either up to 4096 16-bit integers if in array format, or a 8kB 16-bit bitmap if in bitmap format. (and have a paper backing up the design with tests)

My naïve extension to 64bit was to extend the key and value size of the containers to 32bits and keep the same guarantee that at the switchover point the size of the integer list is the same as the size of the full bitmap in the container. This meant the 64-bit space was divided into ~4 billion (2^32) containers; each containing either up to ~134 million (2^27) 32-bit integers if in array format, or a 512MB 32-bit bitmap if in bitmap format. (and _do not_ have a paper backing up the design with tests)

Those numbers are a bit large for current memory sizes. There's probably a few workloads that the scheme would work for, but in most cases I think a sorted list and binary search would work better as it doesn't have the overhead of checking for the bitmap format you're unlikely to reach.

---

In terms of adding a 64-bit version I would definitely be for it, there's some information on better ways to implement it at https://github.com/RoaringBitmap/RoaringBitmap/issues/109 and https://github.com/RoaringBitmap/CRoaring/issues/1, and the C++ layer of `CRoaring` has an implementation of the TreeMap method described in http://r-libre.teluq.ca/930/1/Roaring64bits.pdf. I just skimmed the paper via Google Translate and it should be pretty simple to implement that method using `std::collections::BTreeMap`.

Nemo157:
Judging by how little coding I've been doing at home lately I doubt I'll get round to working on this myself anytime soon, if anyone feels like working on it I would be happy to offer whatever help is needed.

bvinc:
I think I can handle it.  I can't read that paper, but I read the C++ code.

inherent.rs has been pretty easy so far.  Let me know if you have any comments.

https://github.com/bvinc/roaring-rs/blob/roaring64/src/roaring64/inherent.rs

Nemo157:
Looks great :) only suggestion I have is maybe call it `RoaringTreemap` or something similar so it's obvious what algorithm is being used. That way if any of the other 64bit algorithms get implemented it's easier for users to choose the one that works best for their dataset.

PR #39, opened 2017-03-09
RoaringTreemap for a 64bit bitmap

bvinc:
Issue #38.  I just finished all of the operations I think the RoaringTreemap should have.  Here are my thoughts about all of the changes:

- Do you like the current module setup where it's in its own module and pulled in with `pub use` in lib.rs?
- Serialization isn't implemented.  I assume it's only defined for 32bit.
- The only thing messy and difficult was iteration.  It took me an hour to figure out that I needed `as _` to coerce a function item into a function pointer.
- Some of the ops really needed for me to be able to delete RoaringBitmaps from the tree as I'm iterating through the tree.  Obviously, rust doesn't like this, but less obviously, there is literally no way to do it with BTreeMap.  I decided to collect keys and make the deletion of entries a post-processing step at the end of the operation.  It sucks that the operation takes O(n) space when it doesn't have to, but compared to the actual bitmap size and the space savings after deleting, it's fine.
- The unit tests are kind of a mess.  Creating a subdirectory under /tests doesn't work.  I just duplicated the RoaringBitmap tests for now and they all pass.  They could probably use some more u64-oriented test data.


Nemo157:
Awesome work 😄 

I've put a couple small issues I noticed going over the source.

> * Do you like the current module setup where it's in its own module and pulled in with pub use in lib.rs?

Yep, that looks like a great way to keep it separated. After merging this I may do the same for `RoaringBitmap` now that there's more than the one struct in this crate.

> * Serialization isn't implemented. I assume it's only defined for 32bit.

I didn't see any mention of it when I was implementing the 32bit version, so I'm fine with not supporting it till it's standardised.

> * The only thing messy and difficult was iteration. It took me an hour to figure out that I needed as _ to coerce a function item into a function pointer.

Uhhhh, yeah, I've spend way too long staring at that error message and not knowing what `rustc` wants me to do. I hope at some point that gets improved.

> * Some of the ops really needed for me to be able to delete RoaringBitmaps from the tree as I'm iterating through the tree. Obviously, rust doesn't like this, but less obviously, there is literally no way to do it with BTreeMap. I decided to collect keys and make the deletion of entries a post-processing step at the end of the operation. It sucks that the operation takes O(n) space when it doesn't have to, but compared to the actual bitmap size and the space savings after deleting, it's fine.

Yep, only other way I could see to implement it is to do something like `.into_iter().filter_map(|| do the operation).collect::<BTreeMap>()` which would almost certainly be less performant. The vector allocation should really not matter compared to how long the actual operation will take.

> * The unit tests are kind of a mess. Creating a subdirectory under /tests doesn't work. I just duplicated the RoaringBitmap tests for now and they all pass. They could probably use some more u64-oriented test data.

Huh, I guess I've never tried using subdirectories before; since that doesn't work having a `treemap_` prefix looks fine to me. I think having some tests that cover the removal of bitmaps during the ops would be good, I guess you can't test that directly, but you can at least set it up so the bitmap is removed directly instead of calling into it's operators and check the values aren't still there afterwards.

bvinc:
> I've put a couple small issues I noticed going over the source.

If you meant to add some comments on the diff, I don't see them.

Nemo157:
Ah, I need to "Request Changes" now, I think this is the first time I've used the new review system.

src/roaring_treemap/iter.rs
```diff
@@ -0,0 +1,181 @@
...
+                                   fn((u32, RoaringBitmap)) -> To64IntoIter>;
+
+/// An iterator for `RoaringTreemap`.
+pub struct Iter<'a> {
+    inner: InnerIter<'a>,
+    size_hint: u64,
+}
+
+/// An iterator for `RoaringTreemap`.
+pub struct IntoIter {
```
bvinc: I think you're right.  I've decided to expose them at roaring_treemap::{Iter, IntoIter} by putting a `pub use` in roaring_treemap/mod.rs.  That way, it follows the model of std::collections::BTreeMap and std::collections::btree_map::Iter.  Does that sound good?  Maybe I should just keep RoaringTreemap in the module as well?

Don't you currently have the same problem with your IntoIter for RoaringBitmap?  I don't see any way it's exposed, but you're using it when you `impl IntoIterator for RoaringBitmap`.  I don't understand how that would work or how the documentation would work.

Ok, I just ran cargo doc, and yeah, it doesn't work.

Nemo157:
Great, I'll make some time tomorrow to stick the existing code into a subdirectory as well, fix the iters not being exposed, then release a new version 🎉 (just realised I hadn't released a new version with my other changes yet, these are gonna be some hefty release notes).

PR #42, opened 2017-04-06
Add serialized_size(), bitmaps() and from_bitmaps()

arthurprs:
Add serialized_size() is a great addition, we can bikeshed the name though. Constructing internals isn't terribly useful, but I really need it to manually serialize it (as there's no standard).

### DIFF
--- a/src/bitmap/serialization.rs
+++ b/src/bitmap/serialization.rs
@@ -11,6 +11,24 @@ const SERIAL_COOKIE: u16 = 12347;
 // const NO_OFFSET_THRESHOLD: u8 = 4;
 
 impl RoaringBitmap {
+    /// Return the size in bytes of the serialized output.
+    /// This is compatible with the official C/C++, Java and Go implementations.
+    pub fn serialized_size(&self) -> usize {
+        let container_sizes: usize = self.containers.iter().map(|container| {
+            match container.store {
+                Store::Array(ref values) => {
+                    8 + values.len() * 2
+                }
+                Store::Bitmap(..) => {
+                    8 + 8 * 1024
+                }
+            }
+        }).sum();
+
+        // header + container sizes
+        8 + container_sizes
+    }
+
     /// Serialize this bitmap into [the standard Roaring on-disk format][format].
     /// This is compatible with the official C/C++, Java and Go implementations.
     ///
--- a/src/treemap/iter.rs
+++ b/src/treemap/iter.rs
@@ -144,6 +144,46 @@ impl RoaringTreemap {
     pub fn iter(&self) -> Iter {
         Iter::new(&self.map)
     }
+
+    /// Iterator over pairs of partition number and the corresponding RoaringBitmap.
+    /// The partition number is defined by the 32 most significant bits of the bit index.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use roaring::{RoaringBitmap, RoaringTreemap};
+    /// use std::iter::FromIterator;
+    ///
+    /// let original = RoaringTreemap::from_iter(0..6000);
+    /// let mut bitmaps = original.bitmaps();
+    ///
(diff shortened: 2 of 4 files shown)


Nemo157:
`serialized_size` is a great idea 😄 I'm relatively happy with the name but wonder if working `bytes` into it somehow could be useful for self-documentation. The equivalents in the other libraries are:
 * [CRoaring `roaring_bitmap_portable_size_in_bytes`](https://github.com/RoaringBitmap/CRoaring/blob/master/include/roaring/roaring.h#L337)
 * [Java Roaring `serializedSizeInBytes`](https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L2095)

`serialized_size_in_bytes` doesn't seem too Rustic though 😟, I'm pretty happy with either.

---

I sort of dislike leaking the fact that `RoaringTreemap`'s using a `BTreeMap` internally in case there's some more efficient TreeMap that this could use in the future (even though it's mentioned in the documentation...). Having `From`/`Into` implementations also reduces the chance users would read the documentation of them compared to inherent methods; and (while it's unlikely a user would use the functions with data of the wrong shape) I feel that documenting what the `u32` keys mean is very important (especially for converting `BTreeMap` -> `RoaringTreemap`) so users don't end up with broken `RoaringTreemap`s.

What do you think of an API like

```rust
struct BitmapIter<'a> { ... }

impl RoaringTreemap {
    pub fn iter_bitmaps(&self) -> BitmapIter { ... }
    pub fn from_bitmap_iter<I: Iterator<Item=(u32, RoaringBitmap)>>(iter: I) -> RoaringTreemap { ... }
}

impl<'a> Iterator for BitmapIter<'a> {
    type Item = (u32, &'a RoaringBitmap);
}
```

This would be slightly more expensive if the user actually had a `BTreeMap` that they want to convert into a `RoaringTreemap`. But I would assume in most cases it would be relatively easy to instead create an iterator directly from where the data is being loaded/constructed and pass that into the method.

arthurprs:
I look `serialized_size()` from bincode, it's probably clear enough.

I took your suggestion but adjusted the naming slight as we don't see the iter prefix/suffix around, we have for example keys(), values(), chars(), from_chars(), etc..

What do you think about using Cow to allow reconstructing from both owned and reference bitmaps?

Nemo157:
Those are definitely some better names :+1:

---

I always go back and forth on `Cow`, it's nice to make stuff more generically usable, but has some pretty high overhead in terms of method signatures. In this case while it might be useful sometimes I'd prefer the user to perform the cloning themselves.

```rust
pub fn from_bitmaps<I>(iterator: I) -> Self
    where I : IntoIterator<Item = (u32, RoaringBitmap)>;

pub fn from_bitmaps<'a, T, I>(iterator: I) -> Self
    where T: Into<Cow<'a, RoaringBitmap>>, I : IntoIterator<Item = (u32, T)>;
```

The first seems a lot nicer to read in the docs.

```rust
let bitmaps: Iterator<u32, &RoaringBitmap> = ...;

RoaringTreemap::from_bitmaps(bitmaps.map(|(i, b)| (i, b.clone())))
RoaringTreemap::from_bitmaps(bitmaps)
```

The first is definitely worse than the second. But it's the sort of code I've become relatively used to seeing in Rust and is very straightforward to write and understand what it's doing. The second also hides the cost of the operation more, having to put that explicit clone there might influence the user towards checking if they can instead passed an iterator of owned bitmaps in without having to change anything else.

---

If you feel like adding some examples to the docs they'd be appreciated, but don't worry if you can't, I'll just add them afterwards :smile:

arthurprs:
.cloned() doesn't work here :/,  but overall your proposal is best. I also added the examples as requested.

Nemo157:
Awesome work, thanks a lot 😀

Nemo157:
Released in 0.5.1 (once travis finishes building it) 🎉

PR #44, opened 2017-04-08
Add remove_range()

arthurprs:
This turned out to be way harder than I though but it's the last missing piece I need. The interface isn't great but accepting .. and ... ranges would require nightly. We can fix once it's in stable.

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -59,3 +59,15 @@ fn is_subset_4(b: &mut Bencher) {
     let sup: RoaringBitmap = (0..65536).collect();
     b.iter(|| test::black_box(sub.is_subset(&sup)))
 }
+
+#[bench]
+fn remove_range_bitmap(b: &mut Bencher) {
+    let mut sub: RoaringBitmap = (0..65536).collect();
+    b.iter(|| {
+        // carefully delete part of the bitmap
+        // only the first iteration will actually change something
+        // but the runtime remains identical afterwards
+        test::black_box(sub.remove_range(4096 + 1..65536));
+        assert_eq!(sub.len(), 4096 + 1);
+    });
+}
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -36,10 +36,72 @@ fn smoke() {
 }
 
 #[test]
+fn remove_range() {
+    let ranges = [0u32, 1, 63, 64, 65, 100, 4096 - 1, 4096, 4096 + 1, 65536 - 1, 65536, 65536 + 1];
+    for (i, &a) in ranges.iter().enumerate() {
+        for &b in &ranges[i..] {
+            let mut bitmap = RoaringBitmap::from_iter(0..65536 + 1);
+            assert_eq!(bitmap.remove_range(a as u64..b as u64), (b - a) as u64);
+            assert_eq!(bitmap, RoaringBitmap::from_iter((0..a).chain(b..65536 + 1)));
+        }
+    }
+}
+
+#[test]
+fn remove_range_array() {
+    let mut bitmap = RoaringBitmap::from_iter(0..1000);
+    for i in 0..1000 {
+        assert_eq!(bitmap.remove_range(i..i), 0);
+        assert_eq!(bitmap.remove_range(i..i + 1), 1);
+    }
+
+    // insert 0, 2, 4, ..
+    // remove [0, 2), [2, 4), ..
+    let mut bitmap = RoaringBitmap::from_iter((0..1000).map(|x| x * 2));
+    for i in 0..1000 {
+        assert_eq!(bitmap.remove_range(i * 2..(i + 1) * 2), 1);
+    }
+
+    // remove [0, 2), [2, 4), ..
+    let mut bitmap = RoaringBitmap::from_iter(0..1000);
+    for i in 0..1000 / 2 {
+        assert_eq!(bitmap.remove_range(i * 2..(i + 1) * 2), 2);
+    }
+}
+
+#[test]
+fn remove_range_bitmap() {
+    let mut bitmap = RoaringBitmap::from_iter(0..4096 + 1000);
+    for i in 0..1000 {
+        assert_eq!(bitmap.remove_range(i..i), 0);
+        assert_eq!(bitmap.remove_range(i..i + 1), 1);
+    }
+
+    // insert 0, 2, 4, ..
+    // remove [0, 2), [2, 4), ..
+    let mut bitmap = RoaringBitmap::from_iter((0..4096 + 1000).map(|x| x * 2));
+    for i in 0..1000 {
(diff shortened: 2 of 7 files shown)


Nemo157:
What about taking in a `std::ops::Range` for now so users can call it as `bitmap.remove_range(5..7)`; it'll be missing `..`, `a..`, `..b`, and `a...b` support, but as you say that can be fixed once `RangeArgument` is stabilised (and being a breaking change then is fine as bumping the minimum Rust version to get the stable `RangeArgument` is a breaking change anyway).

Nemo157:
I've done a first pass over this and so far it looks great 😄. I need to take another look at the implementation for bitmaps since that's some pretty dense code.

Looks like `clippy` has picked up a couple of iterator related simplifications as well, copying them in here for ease of reading

```
warning: the loop variable `key` is only used to index `bits`., #[warn(needless_range_loop)] on by default
   --> src/bitmap/store.rs:100:17
    |
100 |                   for key in start_key + 1..end_key {
    |  _________________^ starting here...
101 | |                     removed += bits[key].count_ones();
102 | |                     // When popcnt is available zeroing in this loop is faster,
103 | |                     // but we opt to perform reasonably on most cpus by zeroing after.
104 | |                     // By doing that the compiler uses simd to count ones.
105 | |                 }
    | |_________________^ ...ending here
    |
help: consider using an iterator
    |                 for <item> in bits.iter().take(end_key).skip(start_key + 1) {
    = help: for further information visit https://github.com/Manishearth/rust-clippy/wiki#needless_range_loop

warning: the loop variable `key` is only used to index `bits`., #[warn(needless_range_loop)] on by default
   --> src/bitmap/store.rs:107:17
    |
107 |                   for key in start_key + 1..end_key {
    |  _________________^ starting here...
108 | |                     bits[key] = 0;
109 | |                 }
    | |_________________^ ...ending here
    |
help: consider using an iterator
    |                 for <item> in bits.iter().take(end_key).skip(start_key + 1) {
    = help: for further information visit https://github.com/Manishearth/rust-clippy/wiki#needless_range_loop

warning: called `filter(p).next()` on an `Iterator`. This is more succinctly expressed by calling `.find(p)` instead., #[warn(filter_next)] on by default
   --> src/treemap/inherent.rs:221:9
    |
221 |           self.map
    |  _________^ starting here...
222 | |             .iter()
223 | |             .filter(|&(_, rb)| rb.min().is_some())
224 | |             .next()
    | |___________________^ ...ending here
    |
    = note: replace `filter(|&(_, rb)| rb.min().is_some()).next()` with `find(|&(_, rb)| rb.min().is_some())`
    = help: for further information visit https://github.com/Manishearth/rust-clippy/wiki#filter_next

warning: called `filter(p).next()` on an `Iterator`. This is more succinctly expressed by calling `.find(p)` instead., #[warn(filter_next)] on by default
   --> src/treemap/inherent.rs:244:9
    |
244 |           self.map
    |  _________^ starting here...
245 | |             .iter()
246 | |             .rev()
247 | |             .filter(|&(_, rb)| rb.max().is_some())
248 | |             .next()
    | |___________________^ ...ending here
    |
    = note: replace `filter(|&(_, rb)| rb.max().is_some()).next()` with `find(|&(_, rb)| rb.max().is_some())`
    = help: for further information visit https://github.com/Manishearth/rust-clippy/wiki#filter_next
```

If you want to run `clippy` locally then you can get the exact version running in CI with

```
rustup install nightly-2016-12-19
cargo +nightly-2016-12-19 install clippy --vers 0.0.104
cargo clippy
```

although I doubt those warnings will have changed if you've got a newer version.

arthurprs:
Nice, I didn't notice that one was in stable already. I kept the internal stuff using 2 indexes to reduce complexity.

arthurprs:
For some reason travis won't show me the log of the failed task.

tests/lib.rs
```diff
@@ -36,10 +36,71 @@ fn smoke() {
 }
 
 #[test]
+fn remove_range() {
+    let ranges = [0, 1, 63, 64, 65, 100, 4096 - 1, 4096, 4096 + 1, 65536 - 1, 65536, 65536 + 1];
+    for (i, &a) in ranges.iter().enumerate() {
+        for &b in &ranges[i..] {
+            let mut bitmap = RoaringBitmap::from_iter(0..65536 + 1);
+            assert_eq!(bitmap.remove_range(a..b), (b - a) as u64);
```
Nemo157: Could you add an assertion here that the bitmap contains the correct values, something like

```rust
assert_eq!(bitmap, RoaringBitmap::from_iter((0..a).chain(b..65536 + 1)));
```

to make sure it's the correct values being removed and there's no off by one or something in the removal.

benches/lib.rs
```diff
@@ -59,3 +59,12 @@ fn is_subset_4(b: &mut Bencher) {
     let sup: RoaringBitmap = (0..65536).collect();
     b.iter(|| test::black_box(sub.is_subset(&sup)))
 }
+
+#[bench]
+fn remove_range_bitmap(b: &mut Bencher) {
+    let mut sub: RoaringBitmap = (0..65536).collect();
+    b.iter(|| {
+        test::black_box(sub.remove_range(4096 + 1, 65536));
```
Nemo157: This closure is going to be called many times for each `sub` that is created, so only the first call will actually remove values from it. I think the benchmark will have to clone the bitmap every time, hopefully the time for cloning is enough less than the time for removing a range to make the benchmark still useful.

tests/lib.rs
arthurprs: Sure, good idea.

benches/lib.rs
arthurprs: It's intentional, to get a precise measurement of the bitmap remove_range.

Nemo157:
> For some reason travis won't show me the log of the failed task.

Me neither, sounds like travis had some database issues saving logs recently so may not be fully reliable at the moment. Running `clippy` locally gave me:

```
error[E0061]: this function takes 1 parameter but 2 parameters were supplied
  --> benches/lib.rs:73:34
   |
73 |         sub.clone().remove_range(4096 + 1, 65536)
   |                                  ^^^^^^^^^^^^^^^ expected 1 parameter
```

benches/lib.rs
Nemo157: Ah, I guess it's still running that code since the bitmap isn't fully removed, just not removing anything each time. Might pay to put a comment here about that.

Nemo157:
So turns out trying to manually merge stuff on github with required CI checks makes stuff really complicated. I merged in your change from before you added the comment, now I've got a rebased change just adding the comment I'll merge, but there's probably no way to link the merge up with this PR for github, so I'm just going to have to close it here.

Thanks a lot for this, I'll push 0.5.2 out with this straight away 😀

PR #46, opened 2019-05-05
Bump byteorder version to 1.x

messense:


messense:
Is it OK to bump minimum rustc version?

Nemo157:
If you're happy with fixing all the warnings then I'll accept updating the minimum version to the latest (in that case you can use the latest stable version to run clippy instead of running it on nightly). Otherwise you can get rid of the later commits, I'll just override CI to merge the version bump and sort out the nightly build later.

messense:
I'd like to fix all the warnings.

messense:
All warnings are fixed, bumped minimum rustc version to 1.17.0

messense:
Ping.

Nemo157:
Sorry I forgot about this. Thanks for doing it, I'll try and release a new version with this soon.

PR #49, opened 2019-08-30
Remove unnecessary lifetime bound

zyctree:
The lifetime bound seems unnecessary


Nemo157:
It would appear you're right, thanks!

bors: r+

PR #51, opened 2019-11-24
Updated to 2018 edition

nassor:
- replaced deprecated `try!` in favor of `?`
- updated modules imports
- improved benchmarks using criterion crate.


Nemo157:
LGTM, thanks!

bors r+

PR #55, opened 2020-03-24
Format code using cargo fmt

josephglanville:
This PR formats the existing code with `cargo fmt` and then enables a check in CI to ensure it remains formatted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,10 @@ matrix:
     rust: stable
     install:
     - rustup component add clippy
+    - rustup component add rustfmt
     script:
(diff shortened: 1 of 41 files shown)


josephglanville:
Feel free to close this if you don't want to use `cargo fmt`. My editor automatically runs `rustfmt` on save so to avoid creating unrelated churn in future PRs this would be an ideal but not necessarily required change.

Nemo157:
:+1: I'm a big fan of `rustfmt`, this repo just predates when it was really usable and I never got round to adding it 😄

bors r+

PR #56, opened 2020-04-02
WIP: Run container

josephglanville:
This is a work in progress PR for run container support.

Currently completed is basic representation of run containers, serde and optimize, i.e creation of run containers from bitmap and array containers when it makes sense to do so.
Also implemented is simple insert/remove, etc.
Operators are not currently implemented and are stubbed out.

Closes #12

Issue #12, opened 2015-09-19
Add support for run containers


lemire:
As of version 0.5, the java version of Roaring bitmaps supports run containers...

https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RunContainer.java

These can be created by the "runOptimize" function...

https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L1286

These containers can improve the compression ratio in some cases.

josephglanville:
I would like to take this on, run containers provide very good wins for my workload.

I propose adding a new `Vec` based container that stores intervals, either as pair tuples or as a struct:
```
struct Interval {
  start: u16,
  end: u16,
}

Run(Vec<Interval>)
```

Not 100% sure (will verify) but I think the struct takes the same space as the tuple `(u16, u16)`.

I think storing `start`, `end` will result in slightly nicer code vs storing `start`, `run_length`.
Serialisation will of course follow the format specification.

Nemo157:
@josephglanville sounds good, let me know if you have any questions.

saik0:
Picking this up.

I'm not clear what equality / cmp semantic we want for roaring bitmaps.

What do the other lang roaring libs do?

```rust
let input: RoaringBitmap = (0..100).collect();

// input it not run compressed
assert!(!input.remove_run_compression())

let compressed = input.clone().run_compress();

// Should this assertion succeed or panic?
assert!(input == compressed)
```

saik0:
Another problem I'm running into when implementing this:

Most of the integration tests make assumptions about the implementation details of roaring's containers. The optimal container for _all_ of the following are run containers. Adding run containers is going to require a large refactoring of our existing tests.

`tests/union_with.rs`
```rust
#[test]
fn array_and_bitmap() {
    let mut bitmap1 = (0..2000).collect::<RoaringBitmap>();
    let bitmap2 = (1000..8000).collect::<RoaringBitmap>();
    let bitmap3 = (0..8000).collect::<RoaringBitmap>();

    bitmap1 |= bitmap2;

    assert_eq!(bitmap1, bitmap3);
}
```

Kerollmops:
Do you think that it could be a good way to fix that by introducing an iterator type that wraps another iterator i.e. `100..10000` with some density parameter i.e. `Density::dense(0..1000)`, `Density::sparse(0..1000)`?

[...]

PR #66, opened 2020-09-12
Run containers part 2

Kerollmops:
I started working on the run containers implementation, my starting point is #56 by @josephglanville.

I fixed some bugs but I'm sure I introduced new ones with the operations on runs, I need to add a lot of new tests for runs.


josephglanville:
Awesome! I don't have much time to help atm but will review and comment where I can. :)

src/bitmap/store.rs
```diff
@@ -300,11 +300,7 @@ impl Store {
                 Array(vec)
             }
             Run(ref intervals) => Array(
-                intervals
-                    .iter()
-                    .map(|iv| iv.start..iv.end)
-                    .flatten()
-                    .collect(),
+                intervals.iter().flat_map(|iv| iv.start..=iv.end).collect()
```
josephglanville, line 302: Doh. Good catch.

josephglanville:
I'm going to start picking this back up @Kerollmops. I noticed the `remove_range` implementation is most likely not working yet which is preventing the new tests I have added from passing so I will focus on fixing that up and adding the remaining run container store operations. I will open a PR against your branch once I have some progress to show.

Kerollmops:
Thanks a lot @josephglanville, indeed it seems like I didn't take enough time to make this method work. I hope you fix it, and wait for your PR, thank you 😄

saik0:
@josephglanville I was taking a look through history in #12 

> I think storing start, end will result in slightly nicer code vs storing start, run_length.
Serialisation will of course follow the format specification

I haven't made an opportunity to look through the code for this PR yet. Is that what were doing?

My concern is if we deviate we wont be able to use the same "frozen" format for lazy containers

gcxfd:
when use insert_range in a empty RoaringBitmap , maybe can use RunContainer ?

josephglanville:
@saik0 yeah that is a valid concern and not something I had thought of at the time but think it makes sense as to make it easier to support mmaped bitmaps without needing separate code for them. I don't think the code is too hard to modify to use start + len instead just have to adjust when searching for intervals etc.

If you want to work on this go ahead, my progress has been slow because I am (as usual) occupied with other things sadly.

tgruben:
I'm unclear what's needed to finish this, is there anything I can do to help?

PR #74, opened 2020-11-01
Add remove_range benchmark that triggers bitmap to array conversion

josephglanville:
Wrote this to verify assumptions of #68 there are other benches that trigger bitmap to array conversion but this nicely isolates it.


Kerollmops:
Thank you very much, as this is a bench I assume it will be triggered on a `cargo test`.
bors r+

Kerollmops (approved):
bors r+

PR #76, opened 2020-11-29
feat: implement optimised insert_range

domodwyer:
Implements an optimised `insert_range()` function for efficiently adding
consecutive elements to the set.

Previously the easiest way to do this was with `extends()` or calling `insert()`
in a loop. To insert 100,000 elements this took about `1,000us` - with the 
`insert_range()` this completes in `~2us`.

<details><summary>
Bechmarks
</summary>

`insert_range()` vs. `insert()` in a loop inserting 0..N:

```text
add_range/roaring/10    time:   [191.80 ns 194.15 ns 197.16 ns]
                        thrpt:  [50.720 Melem/s 51.506 Melem/s 52.137 Melem/s]
                 change:
                        time:   [-58.031% -57.320% -56.475%] (p = 0.00 < 0.05)
                        thrpt:  [+129.75% +134.30% +138.27%]
                        Performance has improved.
						
add_range/pre_populated_roaring/10
                        time:   [177.77 ns 180.03 ns 182.21 ns]
                        thrpt:  [54.881 Melem/s 55.547 Melem/s 56.252 Melem/s]
                 change:
                        time:   [-36.465% -34.812% -33.125%] (p = 0.00 < 0.05)
                        thrpt:  [+49.532% +53.404% +57.393%]
                        Performance has improved.

add_range/roaring/100   time:   [342.94 ns 346.12 ns 351.19 ns]
                        thrpt:  [284.75 Melem/s 288.92 Melem/s 291.60 Melem/s]
                 change:
                        time:   [-85.032% -84.862% -84.654%] (p = 0.00 < 0.05)
                        thrpt:  [+551.63% +560.58% +568.10%]
                        Performance has improved.

add_range/pre_populated_roaring/100
                        time:   [366.68 ns 371.57 ns 376.29 ns]
                        thrpt:  [265.76 Melem/s 269.13 Melem/s 272.72 Melem/s]
                 change:
                        time:   [-83.662% -83.417% -83.156%] (p = 0.00 < 0.05)
                        thrpt:  [+493.70% +503.04% +512.08%]
                        Performance has improved.

add_range/roaring/1000  time:   [1.6535 us 1.6579 us 1.6630 us]
                        thrpt:  [601.33 Melem/s 603.16 Melem/s 604.78 Melem/s]
                 change:
                        time:   [-91.569% -91.513% -91.457%] (p = 0.00 < 0.05)
                        thrpt:  [+1070.6% +1078.3% +1086.1%]
                        Performance has improved.

[...]

PR #77, opened 2020-12-04
Allow using from_iter instead of collect

Kerollmops:
Clippy started warning that we use from_iter instead of collect, but why not just letting us do that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,4 +42,4 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets -- -D warnings
+          args: --all-targets
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::from_iter_instead_of_collect)]
+
 extern crate roaring;
 
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,7 +9,7 @@
 
 #![warn(missing_docs)]
 #![warn(variant_size_differences)]
-#![allow(unknown_lints)] // For clippy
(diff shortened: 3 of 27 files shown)


Kerollmops:
bors try

.github/workflows/test.yml
```diff
@@ -42,4 +42,4 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets -- -D warnings
+          args: --all-targets
```
domodwyer, line 45: Do you still need to remove this if we have `#![allow(clippy::from_iter_instead_of_collect)]` in the appropriate places? It's probably best to deny warnings - it prevents things like unused imports, etc.

If adding it back in still causes a failure, maybe try:

```suggestion
          args: --all-targets -- -D warnings -A clippy::from_iter_instead_of_collect
```

domodwyer:
The lint description makes sense: https://rust-lang.github.io/rust-clippy/master/#from_iter_instead_of_collect

I haven't looked but seems like we can probably fix this without a breaking change - I can take a look this too and PR to re-enable the lint over the weekend if it's still a problem 👍

Kerollmops:
@domodwyer If you want to take some time to fix every place where we use the from_iter method where we should use collect, I would be very thankful :)

.github/workflows/test.yml
Kerollmops: The `from_iter_instead_of_collect` lint is an unstable flag and only available on nightly, the problem is that it is complex/impossible to only allow it on nightly.

domodwyer:
Easy! https://github.com/RoaringBitmap/roaring-rs/pull/78

I will rebase #76 after the PR is merged 👍

Kerollmops:
Thank you for #78 but you don't need to rebase, bors will do that for us automatically :)

PR #78, opened 2020-12-04
test: use collect() over from_iter()

domodwyer:
Change all calls to from_iter() into calls to collect().

It was only test code so no worries! Quick bit of regex (sometimes) saves the day 👍 

Closes #77.


Kerollmops (approved):
Yay! Awesome, thank you! I was trying to use the `cargo fix` feature but wasn't working, even if it was showing valid replacements :)

bors r+

PR #83, opened 2021-01-19
Use Vec::retain instead of Vec::remove

Kerollmops:
There are many places where we use [the `Vec::remove` method](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove). It can bring a lot of performances issues as every time this function is called, the right part of the Vec (after the removed element) is shifted to the left by one, when a lot of element must be removed, we should always use [`Vec::retain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain) which is more optimized for this in the sense that it will avoid calling a lot the copy function for when multiple elements must be removed at the same time.

I found that doing this change in the `Store::intersect_with` method for the Array-Array operation gave me much better performances (7m intersected with 3m and intersected with 4m again), moving from 240ms to 55ms. I am sure that we can optimize the current implementation more.

- [x] Stores intersections.
- [x] Stores differences.
- [ ] Stores symmetric differences.

### DIFF
--- a/src/bitmap/cmp.rs
+++ b/src/bitmap/cmp.rs
@@ -118,7 +118,7 @@ impl<'a> Iterator for Pairs<'a> {
             Right,
             Both,
             None,
-        };
+        }
         let which = match (self.0.peek(), self.1.peek()) {
             (None, None) => Which::None,
             (Some(_), None) => Which::Left,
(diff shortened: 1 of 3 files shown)


Kerollmops:
bors try

Kerollmops:
bors try

Kerollmops:
bors try

Kerollmops:
bors try

Kerollmops:
bors try

Kerollmops:
bors try

PR #85, opened 2021-01-22
Replace Vec remove by a retain_mut function in Container operations

Kerollmops:
We introduce a `retain_mut` function that allows us to mutate the iterated element. There already were [a discussion about this `Vec::retain` parameter restriction](https://github.com/rust-lang/rust/issues/25477).

Using the `retain` algorithm is much better when multiple stores must be removed in one batch, same as in #83.


Kerollmops:
bors try

PR #92, opened 2021-03-20
Rework the insert/remove_range functions with RangeBounds

### LINKED ISSUE #5: Implement Bounded Iterator
Waiting on `Bound` to be merged with rust-lang/rust#20082.

[§ describing bounded iterators in the RFC](https://github.com/aturon/rfcs/blob/collections-conventions/text/0000-collection-conventions.md#bounded-iterators)

``` rust
fn range(&self, min: Bound<u32>, max: Bound<u32>) -> RangedItems<'a, u32>;
```

Kerollmops:
This PR fixes #5 by reworking the `Bitmap::insert_range` and `Bitmap::remove_range` functions to accept any type that implement [the `RangeBounds` trait](https://doc.rust-lang.org/nightly/core/ops/trait.RangeBounds.html). Note that it is a breaking change and therefore involves bumping the crate version carefully.

The current version of all these functions was accepting an [exclusive `Range<u64>`](https://doc.rust-lang.org/nightly/core/ops/struct.Range.html) to let user define all possible integers in the range `0` to `u32::MAX`, but as it is an exclusive range, a `u64` was required.

@josephglanville, could you please take a look at this PR? When you got time 😃

### DIFF
--- a/src/bitmap/inherent.rs
+++ b/src/bitmap/inherent.rs
@@ -59,88 +59,117 @@ impl RoaringBitmap {
     /// assert!(!rb.contains(4));
     /// ```
     pub fn insert_range<R: RangeBounds<u32>>(&mut self, range: R) -> u64 {
-        let start = match range.start_bound() {
-            Bound::Included(value) => *value,
-            Bound::Excluded(value) => match value.checked_add(1) {
-                Some(value) => value,
-                None => return 0,
-            },
-            Bound::Unbounded => 0,
+        use Bound::{Included, Excluded, Unbounded};
+        use util::split;
+
+        let range = match (range.start_bound(), range.end_bound()) {
+            (Included(start), Included(end)) => todo!(),
+            (Included(start), Excluded(end)) => todo!(),
+            (Included(start), Unbounded)     => todo!(),
+            (Excluded(start), Included(end)) => todo!(),
+            (Excluded(start), Excluded(end)) => todo!(),
+            (Excluded(start), Unbounded)     => todo!(),
+            (Unbounded,       Included(end)) => todo!(),
+            (Unbounded,       Excluded(end)) => todo!(),
+            (Unbounded,       Unbounded)     => split(0)..=split(u32::max_value()),
         };
 
-        let end = match range.end_bound() {
-            Bound::Included(value) => match value.checked_add(1) {
-                Some(value) => value,
-                None => return 0,
-            },
-            Bound::Excluded(value) => *value,
-            Bound::Unbounded => u32::max_value(),
-        };
+        // let start = match range.start_bound() {
+        //     Bound::Included(value) => util::split(*value),
+        //     Bound::Excluded(value) => util::split(*value),
+        //     Bound::Unbounded => util::split(0),
+        // };
 
-        if end.saturating_sub(start) == 0 {
-            return 0;
-        }
+        // let end = match range.end_bound() {
+        //     Bound::Included(value) => util::split(*value),
+        //     Bound::Excluded(value) => util::split(*value),
+        //     Bound::Unbounded => util::split(u32::max_value()),
+        // };
 
-        let (start_container_key, start_index) = util::split(start);
-        let (end_container_key, end_index) = util::split(end);
+        // dbg!(start, end);
 
-        // Find the container index for start_container_key
-        let start_i = match self.containers.binary_search_by_key(&start_container_key, |c| c.key) {
-            Ok(loc) => loc,
-            Err(loc) => {
-                self.containers.insert(loc, Container::new(start_container_key));
-                loc
-            }
-        };
+        // ...
 
-        // If the end range value is in the same container, just call into
-        // the one container.
-        if start_container_key == end_container_key {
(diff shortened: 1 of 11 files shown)


src/bitmap/inherent.rs
```diff
@@ -64,17 +58,31 @@ impl RoaringBitmap {
...
-            "can't index past 2**32"
-        );
-        if range.is_empty() {
+    pub fn insert_range<R: RangeBounds<u32>>(&mut self, range: R) -> u64 {
+        let start = match range.start_bound() {
+            Bound::Included(value) => *value,
+            Bound::Excluded(value) => match value.checked_add(1) {
+                Some(value) => value,
+                None => return 0,
+            },
```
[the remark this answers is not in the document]
Kerollmops: Nope, because I want to early return :)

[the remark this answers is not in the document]
Urhengulas: Ahh, makes sense.

src/bitmap/inherent.rs
```diff
@@ -64,17 +58,31 @@ impl RoaringBitmap {
...
+            Bound::Unbounded => 0,
+        };
+
+        let end = match range.end_bound() {
+            Bound::Included(value) => match value.checked_add(1) {
+                Some(value) => value,
+                None => return 0,
+            },
```
tomjw64, line 75: I find it a bit strange that of these three hypothetical test cases, the first two pass and the third does not:
```rust
#[test]
fn test_insert_max_u32() {
    let mut b = RoaringBitmap::new();
    let inserted = b.insert(u32::MAX);
     // We are allowed to add u32::MAX
    assert!(inserted);
}

#[test]
fn test_insert_range_zero_inclusive() {
    let mut b = RoaringBitmap::new();
    let inserted = b.insert_range(0..=0);
    // `insert_range(value..=value)` appears equivalent to `insert(value)`
    assert_eq!(inserted, 1);
    assert!(b.contains(0));
}

#[test]
fn test_insert_range_max_u32_inclusive() {
    let mut b = RoaringBitmap::new();
    let inserted = b.insert_range(u32::MAX..=u32::MAX);
     // But not equivalent for u32::MAX
    assert_eq!(inserted, 1); // Fails - left: 0, right: 1
    assert!(b.contains(u32::MAX));
}
```

tomjw64: To add onto this, since it does not seem possible with `insert_range` to add `u32::MAX` (any `Range` argument with an `end_bound` of `Inclusive(u32::MAX)` will insert 0 elements), if this is intended behavior, wouldn't this function be able to return a `u32`?

Kerollmops: Hello 👋 @tomjw64

Hum... Thank you very much, I borrowed your tests and will investigate this bug.
Indeed, this is not correct, we should be able to insert up to `u32::MAX` by using an inclusive bound.

src/bitmap/inherent.rs
```diff
@@ -64,70 +72,52 @@ impl RoaringBitmap {
...
+    where
+        R: RangeBounds<u32>,
+    {
+        let (start, end);
+        if let Some(range) = util::convert_range_to_inclusive(range) {
+            start = *range.start();
+            end = *range.end();
+        } else {
             return 0;
         }
```
oliverdding, line 85: It should be replaced by:

```suggestion
        let (start, end) = match util::convert_range_to_inclusive(range) {
            Some(range) => (*range.start(), *range.end()),
            None => return 0,
        };
```

It's my fault :( I forgot to change this one.

src/treemap/inherent.rs
```diff
@@ -104,31 +101,36 @@ impl RoaringTreemap {
...
-                    removed += rb.len();
-                    keys_to_remove.push(key);
+                let b = if key == end_container_key {
+                    end_index
                 } else {
-                    removed += rb.remove_range(a..b);
-                    if rb.is_empty() {
-                        keys_to_remove.push(key);
-                    }
+                    u32::max_value()
```
oliverdding, line 129: > `max_value` is deprecated since TBD: replaced by the `MAX` associated constant on this type 

I would recommand replacing them in a good time.

src/bitmap/store.rs
```diff
@@ -136,28 +151,36 @@ impl Store {
...
-                let start_bit = bit(start as u16) as u32;
-                // end_key is inclusive
-                let end_key = key((end - 1) as u16) as usize;
-                let end_bit = bit(end as u16) as u32;
+                let (start_key, start_bit) = (key(start), bit(start));
+                let (end_key, end_bit) = (key(end), bit(end));
 
                 if start_key == end_key {
-                    let mask = (!0u64 << start_bit) & (!0u64).wrapping_shr(64 - end_bit);
+                    let mask = (!0u64 << start_bit) & (!0u64 >> (63 - end_bit));
```
oliverdding, line 183: I'm not sure if `!0u64` is necessary, since `!0u64` = `u64::MAX`, and the later one is much more straightforward.

Kerollmops:
Thank you very much for your review @oliverdding, I have applied your advice and will merged just after your last review 🫖

Kerollmops:
bors try

PR #94, opened 2021-04-20
Optimize the BitAnd/Assign and BitOr/Assign operations

Kerollmops:
This is a successful attempt at optimizing the `&`, `&=`, `|`, and `|=` by making sure that we do the operation on the smallest or biggest roaring bitmap depending on the operation.

There is room for a lot more optimization in the union operation (i.e. `&`, `&=`) where the containers that don't exist on one side are cloned from the other side instead of being extracted when the other side is owned and will be dropped anyway.
I have [filled an issue](https://github.com/RoaringBitmap/roaring-rs/issues/95) with my concerns about the future of the `intersect_with`, `union_with`... `RoaringBitmap` methods and the equivalent on the internal `Container`/`Store` structs.

### DIFF
--- a/src/bitmap/ops.rs
+++ b/src/bitmap/ops.rs
@@ -217,47 +217,61 @@ impl<'a> BitOrAssign<&'a RoaringBitmap> for RoaringBitmap {
 }
 
 impl BitAnd<RoaringBitmap> for RoaringBitmap {
-    type Output = crate::RoaringBitmap;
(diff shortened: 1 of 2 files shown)


josephglanville (approved):

Kerollmops:
Thank you @josephglanville for your review!

bors merge

Kerollmops:
bors merge

PR #97, opened 2021-04-22
Replace set ops literal methods (`intersect_with`...) by the Rust std ops traits (`BitAnd`...)

### LINKED ISSUE #95: Replace the internal `Container`/`Store` operation methods by the operation traits
The `union_with`, `intersect_with`... methods shadows a lot of potential optimizations by only accepting a reference of another `Container`/`Store`. Parameters cannot be swapped or altered even if the user could give a borrowed value.

We should replace [the `intersect_with`, `union_with`... methods](https://docs.rs/roaring/0.6.5/roaring/bitmap/struct.RoaringBitmap.html#impl-4) with the `BitAnd`, `BitOr`... traits equivalent where we would be able to support owned-owned operations to speed them up by avoiding a lot of clones. As those methods are not leaked publicly it is not breaking. The [`BitAnd`/`BitOr` traits](https://doc.rust-lang.org/std/ops/trait.BitAnd.html) allow us to define specific operations for owned-owned, ref-owned... operations.

To be able to use the power of these new improved operations on the `Container` and `Store` types we must use them in the `BitAnd`/`BitOr` `RoaringBitmap` traits implementations accordingly. The only remaining thing to do is to mark the raw `intersect_with`, `union_with`... as deprecated, advising the use of the ops methods instead.

Kerollmops:
This pull request fixes #95 by implementing better versions of the set operations (union, intersection, difference, and symmetric difference) by efficiently using [the standard ops traits](https://doc.rust-lang.org/std/ops/trait.BitOr.html) (`BitOr`, `BitAnd`, `Sub`, and `BitXor`) with the concept of ownership. For example, if an operation must be done on two owned types, the library will take the owned internal store instead of cloning them.

In [a previous pull request](https://github.com/RoaringBitmap/roaring-rs/issues/94) we made the first step toward a more efficient version of the `RoaringBitmap` standard ops traits, but it was not applied to the `Container` and `Store` internal types, so the result was good but not enough.

We also decided it was mandatory to deprecate the [raw operation methods](https://docs.rs/roaring/0.6.6/roaring/bitmap/struct.RoaringBitmap.html#impl-4) (`union_with`, `intersect_with`, `difference_with`, `symmetric_difference_with`) as those were only exposing one way to do the operations (mutable with borrowed) when users could sometime provide two owned bitmaps and get performance gains by using [the appropriate ops trait variant](https://docs.rs/roaring/0.6.6/roaring/bitmap/struct.RoaringBitmap.html#impl-BitAnd%3C%26%27_%20RoaringBitmap%3E). The only downside of eventually removing these raw methods is the loss in clarity, for me, at least, it is easier to understand `intersect_with` than `&=` or `BitAndAssign::bitand_assign`.

### DIFF
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -1,4 +1,4 @@
-use std::ops::BitOrAssign;
+use std::ops::{BitAndAssign, BitOrAssign};
 use std::{fmt, ops::Range};
 
 use super::store::{self, Store};
@@ -105,9 +105,7 @@ impl Container {
         note = "Please use the `BitAndAssign::bitand_assign` ops method instead"
     )]
     pub fn intersect_with(&mut self, other: &Self) {
-        self.store.intersect_with(&other.store);
-        self.len = self.store.len();
-        self.ensure_correct_store();
+        BitAndAssign::bitand_assign(self, other)
     }
 
     #[deprecated(
@@ -161,6 +159,22 @@ impl BitOrAssign<Container> for Container {
 impl BitOrAssign<&Container> for Container {
     fn bitor_assign(&mut self, rhs: &Container) {
         BitOrAssign::bitor_assign(&mut self.store, &rhs.store);
+        self.len = self.store.len();
+        self.ensure_correct_store();
+    }
+}
+
+impl BitAndAssign<Container> for Container {
+    fn bitand_assign(&mut self, rhs: Container) {
+        BitAndAssign::bitand_assign(&mut self.store, rhs.store);
+        self.len = self.store.len();
+        self.ensure_correct_store();
+    }
+}
+
+impl BitAndAssign<&Container> for Container {
+    fn bitand_assign(&mut self, rhs: &Container) {
+        BitAndAssign::bitand_assign(&mut self.store, &rhs.store);
         self.len = self.store.len();
         self.ensure_correct_store();
     }
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -91,24 +91,40 @@ impl Container {
         self.len <= other.len && self.store.is_subset(&other.store)
     }
 
+    #[deprecated(
+        since = "0.6.7",
+        note = "Please use the `BitOrAssign::bitor_assign` ops method instead",
+    )]
     pub fn union_with(&mut self, other: &Self) {
         self.store.union_with(&other.store);
         self.len = self.store.len();
         self.ensure_correct_store();
     }
 
+    #[deprecated(
+        since = "0.6.7",
+        note = "Please use the `BitAndAssign::bitand_assign` ops method instead",
+    )]
     pub fn intersect_with(&mut self, other: &Self) {
         self.store.intersect_with(&other.store);
         self.len = self.store.len();
         self.ensure_correct_store();
     }
 
+    #[deprecated(
+        since = "0.6.7",
+        note = "Please use the `SubAssign::sub_assign` ops method instead",
+    )]
     pub fn difference_with(&mut self, other: &Self) {
         self.store.difference_with(&other.store);
         self.len = self.store.len();
         self.ensure_correct_store();
     }
 
+    #[deprecated(
+        since = "0.6.7",
+        note = "Please use the `BitXorAssign::bitxor_assign` ops method instead",
+    )]
     pub fn symmetric_difference_with(&mut self, other: &Self) {
         self.store.symmetric_difference_with(&other.store);
         self.len = self.store.len();
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -1,3 +1,4 @@
+use std::ops::BitOrAssign;
 use std::{fmt, ops::Range};
 
 use super::store::{self, Store};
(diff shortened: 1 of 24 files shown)


Kerollmops:
bors try

Kerollmops:
bors try

Kerollmops:
bors try

Kerollmops:
bors try

Kerollmops:
Hey @josephglanville, do you have time to review this one, please? 😃

josephglanville:
Yes I will take a look tomorrow. :)

Kerollmops:
Added the clippy fixes of @MarinPostma.

bors try

Kerollmops:
bors try

src/treemap/ops.rs
```diff
@@ -170,208 +145,320 @@ impl RoaringTreemap {
...
+        for (key, other_rb) in rhs.map {
+            match self.map.entry(key) {
+                Entry::Vacant(ent) => {
+                    ent.insert(other_rb);
+                }
+                Entry::Occupied(mut ent) => {
+                    BitXorAssign::bitxor_assign(ent.get_mut(), other_rb);
+                    if ent.get().is_empty() {
+                        keys_to_remove.push(key);
+                    }
```
MarinPostma, line 431: why not remove it here with `OccupiedEntry::remove_entry`?

src/treemap/ops.rs
```diff
@@ -170,208 +145,320 @@ impl RoaringTreemap {
...
+                SubAssign::sub_assign(self_rb, other_rb);
+                if self_rb.is_empty() {
+                    keys_to_remove.push(*key);
+                }
+            }
+        }
+
+        for key in keys_to_remove {
+            self.map.remove(&key);
+        }
```
MarinPostma, line 372: you should be able to iterate on the k/v of the rhs, and get the entry from the lhs to remove them on the fly, avoiding using a vec.

MarinPostma (changes requested):
My understanding of roaring bitmap is not deep enough to provide an in depth review of the algorithms, but I have noted a thing or two that may be interesting.

src/bitmap/container.rs
```diff
@@ -91,47 +92,79 @@ impl Container {
...
+            (store @ &Store::Array(..), len) if len > ARRAY_LIMIT => Some(store.to_bitmap()),
+            _ => None,
+        };
+        if let Some(new_store) = new_store {
+            self.store = new_store;
+        }
+    }
+}
+
+impl BitOrAssign<Container> for Container {
```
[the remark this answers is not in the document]
Kerollmops: Please stop, don't use macro everywhere! :)

src/bitmap/container.rs
```diff
@@ -91,47 +92,79 @@ impl Container {
...
         self.len = self.store.len();
         self.ensure_correct_store();
     }
+}
 
-    pub fn intersect_with(&mut self, other: &Self) {
-        self.store.intersect_with(&other.store);
+impl BitOrAssign<&Container> for Container {
+    fn bitor_assign(&mut self, rhs: &Container) {
+        BitOrAssign::bitor_assign(&mut self.store, &rhs.store);
```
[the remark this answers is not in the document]
Kerollmops: I prefer using [the fully qualified syntax to be sure of the method I am using](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name).

src/bitmap/ops.rs
```diff
@@ -154,230 +148,347 @@ impl RoaringBitmap {
...
+    /// This is equivalent to an `union` between both maps.
     fn bitor_assign(&mut self, rhs: &RoaringBitmap) {
-        self.union_with(rhs)
+        for container in &rhs.containers {
+            let key = container.key;
+            match self.containers.binary_search_by_key(&key, |c| c.key) {
+                Err(loc) => self.containers.insert(loc, container.clone()),
+                Ok(loc) => BitOrAssign::bitor_assign(&mut self.containers[loc], container),
+            }
+        }
```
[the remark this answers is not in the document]
Kerollmops: That's a great idea indeed, but I think we should dedicate this to a specific PR and add a benchmark to make sure it as any impact. Thank you :)

src/treemap/ops.rs
Kerollmops: Whoops, thank you. I indeed forget about it!

src/treemap/ops.rs
Kerollmops: That's indeed a lot better, thank you for that!

Kerollmops:
bors try

src/treemap/ops.rs
Kerollmops: BTW this same change could have been done on `BitAnd::bitand` as well but I made sure that the smallest map was iterated on, which made the algorithm faster. However, I am waiting on [the `BtreeMap::retain` method](https://doc.rust-lang.org/std/collections/btree_map/struct.BTreeMap.html#method.retain) to be available (Rust 1.53) to do this loop efficiently (loop on the smallest map) and avoid using a Vec.

MarinPostma (approved):
lgtm

Kerollmops:
Thank you!

bors merge

PR #99, opened 2021-04-28
Non panicking push method

MarinPostma:
Changes the `push` method to make it not panic when inserting a value that is not greater than the previous. Also make some optimizations that make the push method roughly 20% faster than before.

### DIFF
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -52,10 +52,16 @@ impl Container {
         inserted
     }
 
-    pub fn push(&mut self, index: u16) {
+    /// Pushes `index` at the end of the container only if `index` is the new max.
+    ///
+    /// Returns whether the `index` was effectively pushed.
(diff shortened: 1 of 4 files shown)


Kerollmops:
Thank you very much.

bors try

MarinPostma:
sorry, forgot to run fmt. fixed now :)

Kerollmops:
Thank you 🧷,

However, I will not be able to merge it before #97, as this is a breaking PR, and #97 is not.
I would like to release a 0.6.7 for #97 and then a 0.7.0 for this one.

bors try

Kerollmops (changes requested):
Could you please fixup (with only one commit message) the first three commits, please?
Keep the clippy fixes in one separate commit.

Kerollmops:
Added the clippy fix of @MarinPostma.

bros try

Kerollmops (approved):
Thank you, LGTM!

bors merge

PR #107, opened 2021-05-28
Improve benchmarks

Kerollmops:
In this PR we make sure that the benchmarks are also used when testing the crate, we aggregate the `from_sorted_iter` benchmarks into one, and change the "Successive" benchmarks behavior.


Kerollmops:
bors merge

PR #108, opened 2021-05-28
Improve clone

Kerollmops:
In this PR we implement the `Clone::clone_from` method, this way it is faster to clone a `RoaringBit/Treemap` by reusing a previous map if possible.


Kerollmops:
bors merge

PR #109, opened 2021-05-28
Introduce multi operations helper methods

### LINKED ISSUE #57: Set operations for multiple maps at a time

Kerollmops:
Closes #58 and Fixes #57.

image

image


saik0:
I tried the "naive lazy" alg as suggested by @lemire in #58 and it is indeed faster (and simpler) for OR. It also does better than sequential for XOR.

Unsurprisingly it's worse for SUB and AND.

My tests were all using `WIKILEAKS_NOQUOTES_SRT`

My experiments are here https://github.com/saik0/roaring-rs/tree/naive-lazy-ops.

saik0:
Curiosity got the better of me, so I tied some other strategies.

1. Inside the multi-ops there the container access pattern is effectively random. BTreeMap and HashMap keyed by the container key were both improvements. HashMap was especially fast when using the identity hash.
2. Copying can be avoided with COW

Kerollmops:
Unfortunately, you broke the link, next time could you [provide a permanent link](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files), please 😄

I think you can open a draft PR on my own branch maybe, this way I can see the changes you made. It seems like you have already done a lot of work there and you PR outsize mine and opening a PR on main/master is maybe better 😂

saik0:
> Unfortunately, you broke the link, next time could you provide a permanent link, please 😄

Yes. Next time. 😄

> I think you can open a draft PR on my own branch maybe, this way I can see the changes you made. It seems like you have already done a lot of work there and you PR outsize mine and opening a PR on main/master is maybe better 😂

The branch includes several slightly different variations of the same alg. It also checks in the sample datasets into the repo for benchmarking. It's hardly merge ready, but it's helpful for evaluating which strategy to choose.

Here's a PR on my own repo: saik0/roaring-rs#1

> Note that I am impressed by the work you do

Thank you! I've had COVID for the last week and have been feeling awful and isolated. Hearing that felt pretty good! 😄

> make it as fast if not faster than the C roaring library!

Thats a tall order!

saik0:
Here are the benchmarks from my experiments on saik0#1

Screen Shot 2022-01-12 at 11 19 53 AM
Screen Shot 2022-01-12 at 11 20 17 AM

Kerollmops:
Wow, the diff with the multi-or is quite impressive, I don't have a comparison with the multi-xor, though! 😃

Issue #113, opened 2021-08-23
The `insert_range` method does not properly handle boundary condition


oliverdding:
## Progress

* [x] Switch inner parameters to RangeBounds
  * [x] insert_range
  * [x] remove_range
* [x] Fix boundary condition in insert_range/remove_range
* [x] Add more tests

## Experiments

```rust
#[test]
fn insert_range_multi() {
    let mut bitmap = RoaringBitmap::new();
    bitmap.insert_range(0..((1_u64 << 16) + 1));
    println!("{:?}", bitmap);
}

#[test]
fn insert_multi() {
    let mut bitmap = RoaringBitmap::new();
    for i in 0..((1 << 16) + 1) {
        bitmap.insert(i);
    }
    println!("{:?}", bitmap);
}

#[test]
fn insert_range_single() {
    let mut bitmap = RoaringBitmap::new();
    bitmap.insert_range(0..(1_u64 << 16));
    println!("{:?}", bitmap); // would panic here
}

#[test]
fn insert_single() {
    let mut bitmap = RoaringBitmap::new();
    for i in 0..(1 << 16) {
        bitmap.insert(i);
    }
    println!("{:?}", bitmap);
}
```

## single partition insertion

After insert 0..(1_u64 << 16) to the bitmap, we expect only one container(bitmap), whose key is 0, exists in the bitmap.

But actually we got:

![insert_range_single](https://user-images.githubusercontent.com/43624923/130414640-8775f601-aa20-4294-93c2-3fc722b1c7ef.png)

## multiply partition insertion

After insert 0..((1_u64 << 16) + 1) to the bitmap, we expect one container(bitmap), whose key is 0, followed by a container(array), whose key is 1.

But we got:

![insert_range_multi](https://user-images.githubusercontent.com/43624923/130415192-7fd3c201-5bc2-4725-96f5-f3a1d70fb204.png)

Kerollmops:
Hey @oliverdding, have you found the time to work on this? 😃

oliverdding:
Hi @Kerollmops , yes, sure. Maybe I have misunderstood ( I'm waiting for your [PR](https://github.com/RoaringBitmap/roaring-rs/pull/92) on `RangeBound` XP).

Anyway, let's see what I can do now. I can modify inner `insert_range` without breaking outside's `insert_range`, is that right?

Inner function's declaration should be changed, in which case should I use `RangeInclusive ` or `RangeBound` to keep consistency with you?

[...]

PR #115, opened 2021-09-23
Insert/remove range with bounds

oliverdding:
Implimentation of insert_range and remove_range functions that accept RangeBounds instead of Range.

This PR fix [#5] and [#113].

Done:

 * [x] Switch inner parameters to RangeBounds
   * [x] insert_range
   * [x] remove_range
 * [x] Fix boundary condition in insert_range/remove_range
 * [x] Add more test cases
 * [x] switching to RangeBounds of container
   * [x] insert_range
   * [x] remove_range

### DIFF
--- a/src/bitmap/inherent.rs
+++ b/src/bitmap/inherent.rs
@@ -1,4 +1,4 @@
-use std::ops::{Bound, RangeBounds};
+use std::ops::{Bound, RangeBounds, RangeInclusive};
 
 use crate::RoaringBitmap;
 
@@ -44,7 +44,21 @@ impl RoaringBitmap {
         container.insert(index)
     }
 
-    /// Inserts a range of values from the set.
+    /// Search for the specific container by the given key.
+    /// Create a new container if not exist.
+    ///
+    /// Return the index of the target container.
+    fn find_container_by_key(&mut self, key: u16) -> usize {
+        match self.containers.binary_search_by_key(&key, |c| c.key) {
+            Ok(loc) => loc,
+            Err(loc) => {
+                self.containers.insert(loc, Container::new(key));
+                loc
+            }
+        }
+    }
+
+    /// Inserts a range of values.
     /// Returns the number of inserted values.
     ///
     /// # Examples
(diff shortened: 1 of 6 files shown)


src/bitmap/store.rs
```diff
@@ -146,28 +166,36 @@ impl Store {
...
-                let start_bit = bit(start as u16) as u32;
-                // end_key is inclusive
-                let end_key = key((end - 1) as u16) as usize;
-                let end_bit = bit(end as u16) as u32;
+                let (start_key, start_bit) = (key(start), bit(start));
+                let (end_key, end_bit) = (key(end), bit(end));
 
                 if start_key == end_key {
-                    let mask = (!0u64 << start_bit) & (!0u64).wrapping_shr(64 - end_bit);
+                    let mask = (!0u64 << start_bit) & (!0u64 >> (63 - end_bit));
```
Kerollmops, line 198: Why do you remove the `wrapping_shr` here?

src/bitmap/store.rs
```diff
@@ -189,8 +217,8 @@ impl Store {
                     *word = 0;
                 }
                 // end key bits
-                removed += (bits[end_key] & (!0u64).wrapping_shr(64 - end_bit)).count_ones();
-                bits[end_key] &= !(!0u64).wrapping_shr(64 - end_bit);
+                removed += (bits[end_key] & (!0u64 >> (63 - end_bit))).count_ones();
+                bits[end_key] &= !(!0u64 >> (63 - end_bit));
```
Kerollmops, line 221: Same here, why do you remove the `wrapping_shr`?

src/treemap/util.rs
```diff
@@ -8,6 +10,31 @@ pub fn join(high: u32, low: u32) -> u64 {
...
+        Bound::Included(&i) => i,
+        Bound::Excluded(&i) => i - 1,
+        Bound::Unbounded => u64::MAX,
+    };
+    if end < start {
+        return (0..=0, true);
+    }
+    (start..=end, false)
+}
+
```
Kerollmops, line 37: Can't you make it generic for `u32`s and `u64`s, don't do it if you need to introduce a new crate i.e. `num-trait`, I prefer that you duplicate this function for u64s like you have done instead of adding a new dep just for that :)

Kerollmops: Can you please apply the same changes as the other `convert_range_to_inclusive` function, please?

Kerollmops (changes requested):

oliverdding:
Hi @Kerollmops 

Thanks for your reviewing!

Recently I might have no time to code since I'm on my way to travel 👻. But I would come back ASAP.

src/bitmap/store.rs
oliverdding: Let's talk about the `wrapping_shr` in the upper conversation and I would mark the below resolved.

src/bitmap/util.rs
```diff
@@ -12,6 +14,31 @@ pub fn join(high: u16, low: u16) -> u32 {
...
+where
+    R: RangeBounds<u32>,
+{
+    if let Bound::Excluded(0) = range.end_bound() {
+        return (0..=0, true);
+    }
+    let start: u32 = match range.start_bound() {
+        Bound::Included(&i) => i,
+        Bound::Unbounded => 0,
+        _ => panic!("Should never be called (insert_range start with Excluded)"),
```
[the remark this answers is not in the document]
oliverdding: Ok, I have no idea about `saturating_sub`, I would give it a look ;)

src/treemap/inherent.rs
```diff
@@ -113,14 +113,14 @@ impl RoaringTreemap {
...
     {
-        let (range, is_empty) = util::convert_range_to_inclusive(range);
-        if is_empty {
+        let (start, end);
+        if let Some(range) = util::convert_range_to_inclusive(range) {
+            start = *range.start();
+            end = *range.end();
+        } else {
             return 0;
         }
```
Kerollmops, line 122: I would prefer that we use a match here instead, it is easier to read and more rust idiomatic :)
Can you do this at the other places in your changes, please?

```suggestion
        let (start, end) = match util::convert_range_to_inclusive(range) {
            Some(range) => (*range.start(), *range.end()),
            None => return 0,
        };
```

src/bitmap/inherent.rs
```diff
@@ -76,14 +76,14 @@ impl RoaringBitmap {
     where
         R: RangeBounds<u32>,
     {
-        let (range, is_empty) = util::convert_range_to_inclusive(range);
-        if is_empty {
+        let (start, end);
```
Kerollmops, line 79: Can you use a match instead, please?

src/bitmap/inherent.rs
```diff
@@ -207,14 +207,14 @@ impl RoaringBitmap {
...
```
Kerollmops, line 210: Can you use a match instead, please?

src/bitmap/util.rs
```diff
@@ -12,6 +14,31 @@ pub fn join(high: u16, low: u16) -> u32 {
...
+    };
+    let end: u32 = match range.end_bound() {
+        Bound::Included(&i) => i,
+        Bound::Excluded(&i) => i - 1,
```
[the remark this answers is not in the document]
oliverdding: I would like to add a check before the `Bound::Excluded(&i) => i - 1,` to return instead of using `saturating_sub` here.

PR #121, opened 2021-12-13
Make sure the `RetainMut::retain_mut` method is effectively used

Kerollmops:
This PR uses the fully qualified syntax to make sure that we use the `retain_mut` trait method of the `RetainMut` trait and not the one from the `vec_retain_mut` nighty feature [that will be added to the standard library](https://github.com/rust-lang/rust/issues/90829) soon.


Kerollmops:
bors merge

Issue #60, opened 2020-07-15
Will simd be introduced？


TheLudlows:
hello, Will simd be introduced?

saik0:
I took at look at portable SIMD. While it does define the packed SIMD types, some of the instructions used by the algs presented in the paper are target feature specific (such as the sse4.2 [cmpistrm](https://doc.rust-lang.org/core/arch/x86_64/fn._mm_cmpistrm.html`))

saik0:
Another consideration is that the vectorized versions (at least for the two I've looked at, OR and AND) are not safe to use in place.

A nice feature we have that rust makes trivial is the ability to take ownership of another `RoaringBitmap`'s container when the right hand side of an operation is to an owned `RoaringBitmap`. For the context of this discussion I will refer to this as "container recycling"

Vectorization and  container recycling are seemingly at odds. When performing an AND, should we use the faster vectorized alg that must allocate, or recycle the allocation and go with the slower one.

Or is there another option? Thinking out loud: We could keep a thread local buffer around, _always_ use the vectorized version (when it's possible, and given it's the best choice). Then swap the pointers.

saik0:
Ported the vectorized intersection alg from CRoaring, and tried using it for `&=` using the thread local / swap pattern I described. 

Fun. I haven't convinced myself it's a good idea, but it sure is _fast_ 

![Screen Shot 2022-01-19 at 2 04 20 AM](https://user-images.githubusercontent.com/997050/150110100-53d69a6a-c6f6-4792-a3aa-d233d9b86cc7.png)

Given the SIMD alg requires specific x86 features, I'm not at all sure how to go about testing.

Curious what you think @Kerollmops, @Nemo157

saik0:
SIMD stuff in my _very_ messy _very_ experimental branch

https://github.com/saik0/roaring-rs/blob/594cf85a67b640daa91de6af1da395b3e2dc6332/src/bitmap/store/op_vector.rs

Kerollmops:
Wow, is it me or is it way faster than the C equivalent? 🚤

> Given the SIMD alg requires specific x86 features, I'm not at all sure how to go about testing.

This is one of the reasons why I am not sure either, do we really want to bring raw SIMD algorithms into this crate? Can't we rely on [the soon-to-be-released stdSIMD module](https://doc.rust-lang.org/nightly/std/simd/)?

> SIMD stuff in my very messy very experimental branch

[...]

PR #127, opened 2022-01-04
Add scalar optimizations from CRoaring / arXiv:1709.07821 section 3

saik0:
### Purpose

This PR adds some optimizations from CRoaring as outlined in  arXiv:1709.07821 section 3

### Overview 
 * All inserts and removes are now branchless (!in arXiv:1709.0782, in CRoaring)
 * Section 3.1 was already implemented, except for `BitmapIter`. This is covered in #125
 * Implement Array-Bitset aggregates as outlined in section 3.2
   * Also branchless 😎
 * Tracks bitmap cardinality while performing bitmap-bitmap ops
   * This is a deviation from CRoaring, and will need to be benchmarked further before this Draft PR is ready
   * Curious to hear what you think about this @lemire 
 * In order to track bitmap cardinality the len field had to moved into `Store::Bitmap`
   * This is unfortunately a cross cutting change
 * `Store` was quite large (LoC) and had many responsibilities. The largest change in this draft is decomposing `Store` such hat it's field variants are two new container types: each responsible for maintaining their invariants and implementing `ops`
   * `Bitmap8K` keeps track of it's cardinality
   * `SortedU16Vec` maintains its sorting
   * `Store` now only delegates to these containers
   * My hope is that this will be useful when implementing run containers. 🤞
   * Unfortunately so much code was  moved this PR is _HUGE_

### Out of scope
 * Inline ASM for Array-Bitset aggregates
 * Section 4 (explicit SIMD). As noted by the paper authors: The compiler does a decent job of autovectorization, though not as good as hand-tuned

### Notes
 * I attempted to emulate the inline ASM Array-Bitset aggregates by using a mix of unsafe ptr arithmetic and x86-64 intrinsics, hoping to compile to the same instructions. I was unable to get it under 13 instructions per iteration (compared to the papers 5). While it was an improvement, I abandoned the effort in favor of waiting for the `asm!` macro to stabilize. https://github.com/rust-lang/rust/issues/72016


saik0:
Got a benchmark harness with ops over the sample datasets.  [The results](https://gist.github.com/saik0/37ab9f280c7ab68aa256e964013c5aa6) are consistent and very promising!

Kerollmops:
That's quit good indeed.

However, I am not sure about the use of the new `asm!` macro to do inline asm. I would like to keep this crate simple and able to compile on most of the platforms.

Maybe we can make sure that using inline asm doesn't break the build, I had too much trouble compiling [croaring-rs](https://github.com/saulius/croaring-rs) (the wrapper of the C roaring library in Rust) due to a lot of raw SIMD function calls, unavailable on the platforms. Maybe we can use [the std::simd](https://doc.rust-lang.org/nightly/std/simd/index.html) soon to be released crate.

saik0:
> That's quit good indeed 🚀 😄
> 
> However, I am not sure about the use of the new `asm!` macro to do inline asm. I would like to keep this crate simple and able to compile on most of the platforms.
> 
> Maybe we can make sure that using inline asm doesn't break the build, I had too much trouble compiling [croaring-rs](https://github.com/saulius/croaring-rs) (the wrapper of the C roaring library in Rust) due to a lot of raw SIMD function calls, unavailable on the platforms. Maybe we can use [the std::simd](https://doc.rust-lang.org/nightly/std/simd/index.html) soon to be released crate?

Let's have this discussion in #60. This PR does not include any inline asm or explicit SIMD. The perf gains here are in making inserts and remove branchless, and by interleaving the cardinality tracking.

saik0:
Maintaining the len in the bitmap store has the additional micro-optimization of making `sizeof<Container>()` smaller, so binary searching containers should be slightly cache-friendlier.

src/bitmap/bitmap_8k.rs
```diff
@@ -0,0 +1,314 @@
...
+    index as usize / 64
+}
+
+#[inline]
+pub fn bit(index: u16) -> usize {
+    index as usize % 64
+}
+
+#[inline]
+fn op_bitmaps(bits1: &mut Bitmap8K, bits2: &Bitmap8K, op: fn(&mut u64, u64)) {
```
Kerollmops, line 246: Why using an `fn()` here and not letting Rust optimize this maybe better by using an `Fn` trait?
```suggestion
fn op_bitmaps<F: Fn(&mut u64, u64)>(bits1: &mut Bitmap8K, bits2: &Bitmap8K, op: F) {
```

Also, don't you think that using a function that does not modify its parameters would be better optimized? Something like `Fn(u64, u64) -> u64`?

src/bitmap/bitmap_8k.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl BitXorAssign<&SortedU16Vec> for Bitmap8K {
+    fn bitxor_assign(&mut self, rhs: &SortedU16Vec) {
+        let mut len = self.len as i64;
+        for &index in rhs.iter() {
+            let (key, bit) = (key(index), bit(index));
+            let old_w = self.bits[key];
+            let new_w = old_w ^ 1 << bit;
+            len += 1 - 2 * (((1 << bit) & old_w) >> bit) as i64; // +1 or -1
```
Kerollmops, line 309: Ho! So here is one way to remove some conditions, no more jumps for that!

src/bitmap/bitmap_8k.rs
```diff
@@ -0,0 +1,314 @@
+use crate::bitmap::sorted_u16_vec::SortedU16Vec;
+use crate::bitmap::store::Store;
+use std::borrow::Borrow;
+use std::ops::{BitAndAssign, BitOrAssign, BitXorAssign, RangeInclusive, SubAssign};
+
+pub const BITMAP_LENGTH: usize = 1024;
+
+#[derive(Clone)]
+pub struct Bitmap8K {
+    len: u64,
```
Kerollmops, line 10: Just to be sure, here we prefer storing an `u64` instead of an `u16` even if there can be only 65536 values in this container because converting from an `u16` to an `u64` when getting the length is not good for performance, right? And it doesn't that much to use an `u16` for alignment issues, it will always be 2 words on a 64bit machine.

src/bitmap/sorted_u16_vec.rs
```diff
@@ -0,0 +1,508 @@
...
+}
+
+impl SortedU16Vec {
+    pub fn new() -> SortedU16Vec {
+        SortedU16Vec { vec: vec![] }
+    }
+
+    pub fn from_vec(vec: Vec<u16>) -> SortedU16Vec {
+        SortedU16Vec { vec }
+    }
```
Kerollmops, line 19:
```suggestion
    pub fn from_sorted_vec(vec: Vec<u16>) -> SortedU16Vec {
        SortedU16Vec { vec }
    }
```

Kerollmops: I am not sure but don't we want this method to be `unsafe` as it can break the results, and we want the user of this method to be sure that the vector is valid?

src/bitmap/bitmap_8k.rs
```diff
@@ -0,0 +1,314 @@
...
+    bits: Box<[u64; BITMAP_LENGTH]>,
+}
+
+impl Bitmap8K {
+    pub fn new(len: u64, bits: Box<[u64; BITMAP_LENGTH]>) -> Bitmap8K {
+        Bitmap8K { len, bits }
+    }
```
Kerollmops, line 17: Don't we want this method to be `unsafe` to be sure that no rules are broken, and be sure that the user of this method is careful!

Kerollmops (changes requested):
This PR is huge indeed but the work is great! It reduces the time it takes to do most of the operations and also reduces the amount of code in the `bitmap/store.rs` file!

Thank you very much! I don't see any reason not to mark this PR as a candidate for merge!

src/bitmap/bitmap_8k.rs
```diff
@@ -0,0 +1,314 @@
...
+        for word in &self.bits[start_key + 1..end_key] {
+            removed += word.count_ones();
+            // When popcnt is available zeroing in this loop is faster,
+            // but we opt to perform reasonably on most cpus by zeroing after.
+            // By doing that the compiler uses simd to count ones.
+        }
+        // do zeroing outside the loop
+        for word in &mut self.bits[start_key + 1..end_key] {
+            *word = 0;
+        }
```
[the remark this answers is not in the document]
saik0: the implementations of `insert_range` and `remove_range` were only moved. The scope of this PR is big enough 😅

src/bitmap/bitmap_8k.rs
```diff
@@ -0,0 +1,314 @@
...
+                self.move_next();
+                return result;
+            } else {
+                self.move_next();
+            }
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        panic!("Should never be called (roaring::Iter caches the size_hint itself)")
```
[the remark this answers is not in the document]
saik0: > I have no idea. I'm not going to touch it unless I understand it.

src/bitmap/bitmap_8k.rs
saik0: Today I learned `fn` is a function pointer and `Fn` is a monomophizable trait. Yes! Great catch!

src/bitmap/bitmap_8k.rs
saik0: Credit to:
https://github.com/RoaringBitmap/CRoaring/blob/f046613e9b75f787b618e7b62197415ea7c2c879/src/bitset_util.c#L985

src/bitmap/bitmap_8k.rs
saik0: It's a u64 because `store::len` returned u64. Another case of "I didnt want to touch it"

Kerollmops: Yup, and I think that's better like that!

src/bitmap/bitmap_8k.rs
saik0: From my understanding `unsafe` in rust refers specifically to memory safety, not for anything that could violate some invariant.

I take your point though. Lets make this `pub(crate)`

src/bitmap/sorted_u16_vec.rs
saik0: I like it and I love it.

src/bitmap/bitmap_8k.rs
Kerollmops: > From my understanding `unsafe` in rust refers specifically to memory safety, not for anything that could violate some invariant.

I am not sure about that, invariants [can be of two kinds: safety and validity](https://internals.rust-lang.org/t/two-kinds-of-invariants-safety-and-validity/8264). In this case we are talking about validity. What if the len doesn't match what's in the Box (got it?)?

Kerollmops: As I understand this more, we should declare a library function as `unsafe` as soon as it is possible to create an invalid state with the parameters, leading to possible future bugs. It can be called library undefined behavior.

src/bitmap/sorted_u16_vec.rs
saik0: @Kerollmops The invariants of SortedU16Vec are not only sorted, but also unique. I'd welcome a suggestion for a better struct name.

Kerollmops: We could maybe use something like `U16VecSet`? and the method to construct it from a `Vec<u16>` can be `from_vec_unchecked`, this method **must** have indications on how to use it correctly (sorted and dedup).

I remember [an old crate I created with the same requirements](https://docs.rs/sdset/latest/sdset/), I called the type `SdSet`, for sorted and deduplicated.

src/bitmap/bitmap_8k.rs
Kerollmops: We should not make it unsafe, but at list rename it `new_uncheked` and make it private, what do you think?

Kerollmops: In fact, you don't need to make private `pub(crate)` any method on the containers, those are not accessible.

saik0: Honestly I'd like to mark all `Container` `Store` `Bitmap8k` `SortedU16Vec` and so on as `pub(crate)` as end users do not interact with these types

Perhaps we can audit the types with a follow up PR before the next breaking point release.

Kerollmops: `Container`, `Store`, `Bitmap8k`, `SortedU16Vec` are not exposed to the end-users, their modules are not `pub` in lib.rs.

saik0:
@Kerollmops I'm checking invariants with debug assertion in: f29fa21. I think it's sufficient. However: deserialize might panic now for malformed inputs! (but only in debug builds)

src/bitmap/bitmap_8k.rs
saik0: > `Container`, `Store`, `Bitmap8k`, `SortedU16Vec` are not exposed to the end-users, their modules are not `pub` in lib.rs.

Ah, OK. That makes sense.

>  We should not make it unsafe, but at list rename it new_unchecked and make it private, what do you think?

It cant be private anyhow, it's required for deserializing.

src/bitmap/sorted_u16_vec.rs
saik0: c694f54 and c130a9b

Note that the only usages of these unchecked creation are in SortedU16Vec::to_bitmap_store and Bitmap8k::to_array_store

Deserialization now uses the "try" versions, which will return an error in the case of invalid data. Huzzah!

src/bitmap/sorted_u16_vec.rs
```diff
@@ -164,6 +172,53 @@ impl Default for SortedU16Vec {
...
+    type Error = Error;
+
+    fn try_from(value: Vec<u16>) -> Result<Self, Self::Error> {
+        let mut iter = value.iter().enumerate();
+        if let Some((_, mut prev)) = iter.next() {
+            for (i, cur) in iter {
+                match cur.cmp(prev) {
+                    Ordering::Less => return Err(Error { index: i, kind: ErrorKind::OutOfOrder }),
+                    Ordering::Equal => return Err(Error { index: i, kind: ErrorKind::Duplicate }),
+                    Ordering::Greater => {}
```
Kerollmops, line 212: I prefer that we use the unit `()` type instead of an empty scope `{}` which itself return unit `()` 😃

```suggestion
                    Ordering::Greater => (),
```

src/bitmap/sorted_u16_vec.rs
```diff
@@ -164,6 +172,53 @@ impl Default for SortedU16Vec {
...
+                }
+                prev = cur;
+            }
+        }
+
+        Ok(SortedU16Vec { vec: value })
```
Kerollmops, line 218: Can you rename `value` into `vec` ? This maybe I find it clearer 😃 

```suggestion
        Ok(SortedU16Vec { vec })
```

Kerollmops (changes requested):

saik0:
Re: Naming

Bitmap8k and SortedU16Vec are named after what they are. However, if we're being honest with ourselves they are not general purpose types.

We could also name them after what they're for. BitmapStore ArrayStore and possibly rename Store enum to StoreKind or StoreType and move all three into a store module.

I don't have strong opinions on this. I'm just trying to think of what would be least surprising to me if I were seeing the codebase for the first time.

saik0:
Screen Shot 2022-01-11 at 11 02 36 AM

Thankfully checks @ deserialization had marginal performance delta

saik0:
The checks are not free, but I think for deserialization correctness matters more than performance. If this change is controversial I can revert it for this PR and we can have a separate discussion.

Screen Shot 2022-01-11 at 11 18 52 AM

Kerollmops:
Re: Naming

> We could also name them after _what they're for_. `BitmapStore` `ArrayStore` and possibly rename `Store` enum to `StoreKind` or `StoreType` and move all three into a `store` module, or include `Container` and call the mod `container`.

I like the idea, I would just like to keep the `Store` enum name. This is a store that can be of one of two forms. This is not just the kind or type of a store, that's a store!

---

Re: Unchecked Deserialization

I am not sure to understand, what are both of these methods? Have you introduced a new method to deserialize without checking some bounds or other stuff? I am not able to find it on your branch.

saik0:
> Re: Unchecked Deserialization
> 
> ~I am not sure to understand, what are both of these methods? Have you introduced a new method to deserialize without checking some bounds or other stuff? I am not able to find it on your branch.~
> 
> Ho! I understand now, you were trying to use the `try_from` and `from_unchecked` versions of the `BitmapStore` and `ArrayStore`, I got it.

> Also, the timings seem quite low, 2.5ms is very low, which dataset were you using to graph that?

Each point is one of the datasets. The one with the most elements is weather_85 if my memory serves me correctly.

In any case, I don't want to hold up this PR with this discussion, as it's out of scope. I will revert deserialize to the prior behavior. We can discuss elsewhere.

Kerollmops (changes requested):
I like what you've done! Indeed I would like you to remove the unchecked version of the deserialize method. Maybe you can create an issue for this specific issue?

saik0:
> I like what you've done! Indeed I would like you to remove the unchecked version of the deserialize method. Maybe you can create an issue for this specific issue?

OK, I'll open up a followup PR to add data validation to deserialize

Kerollmops:
Do you think I can merge this PR while waiting for your follow-up PR on data validation?

saik0:
> Do you think I can merge this PR while waiting for your follow-up PR on data validation?

Yes. Follow up PR will depend on the try versions implemented in this one, so it's required to merge

Kerollmops:
bors merge

PR #131, opened 2022-01-10
Append perf bugfix

### LINKED ISSUE #130: from_sorted_iter strictly slower than collecting
This one surprised me. It's on the order _10x_ slower.

Screen Shot 2022-01-09 at 9 35 20 PM
Screen Shot 2022-01-09 at 9 35 36 PM

saik0:
Fixes #130 

## Cause

Where _N_ is the length of the iterator: _N_ calls to `RoaringBitmap::push` required N calls to`RoaringBitmap::max`, which is not cheap.

## Solution

Compare only the first item in the iterator to the bitmap max, then compare each element to its predecessor.

## Future work

This could be strictly faster than inserting, given that it's always inserting into the last container (or pushing a new container)

## Before
Screen Shot 2022-01-09 at 9 35 20 PM
Screen Shot 2022-01-09 at 9 35 36 PM

## After

Screen Shot 2022-01-09 at 11 13 35 PM
Screen Shot 2022-01-09 at 11 13 43 PM

### DIFF
--- a/src/bitmap/iter.rs
+++ b/src/bitmap/iter.rs
@@ -184,13 +184,35 @@ impl RoaringBitmap {
         &mut self,
         iterator: I,
     ) -> Result<u64, NonSortedIntegers> {
-        let mut count = 0;
+        // Name shadowed to prevent accidentally referencing the param
+        let mut iterator = iterator.into_iter();
+
+        let mut prev: u32 = match iterator.next() {
+            None => return Ok(0),
+            Some(first) => {
+                if let Some(max) = self.max() {
+                    if first <= max {
+                        return Err(NonSortedIntegers { valid_until: 0 });
+                    }
+                }
+
+                first
+            }
+        };
+
+        self.insert(prev);
+        let mut count = 1;
+
+        // It is now guaranteed that so long as the values are iterator are monotonically
+        // increasing they must also be the greatest in the set.
 
         for value in iterator {
-            if self.push(value) {
-                count += 1;
-            } else {
+            if value <= prev {
                 return Err(NonSortedIntegers { valid_until: count });
+            } else {
+                self.insert(value);
+                prev = value;
+                count += 1;
             }
         }
 
--- a/tests/push.rs
+++ b/tests/push.rs
@@ -26,6 +26,32 @@ fn append() {
 }
 
 #[test]
+fn append_empty() {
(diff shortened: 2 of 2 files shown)


saik0:
As an experiment, removing the binary search was "only" a ~30% improvement. I'm leaving it out as a more general solution is outlined in @richardstartin post under [optimal sequence construction](https://richardstartin.github.io/posts/roaringbitmap-performance-tricks#optimal-sequential-construction) that could also be applied to non-sorted extending the bitmap.

Kerollmops:
I really thought that [`RoaringBitmap::max` was fast enough](https://github.com/RoaringBitmap/roaring-rs/blob/9a1b540bb149f2e83ccf78f819a03cd1120e4615/src/bitmap/inherent.rs#L335-L337) as it is basically just fetching the last container and the last number in it. I also thought that Rust would have optimized that, but was dumb to think that, that's too complex for the optimizer to understand that by itself!

However, thank you very much for that PR, again! Love the energy you put into this crate ❤️

> As an experiment, removing the binary search was "only" a ~30% improvement (above the fix). I'm leaving it out as a more general solution is outlined in @richardstartin post under [optimal sequence construction](https://richardstartin.github.io/posts/roaringbitmap-performance-tricks#optimal-sequential-construction) that could also be applied to unsorted extending the bitmap.

Yeah, 30% (after this optimization) is a lot! How much complexity does it bring? Because implementing the more general solution can take a lot of time too!

Kerollmops:
bors merge

saik0:
> Yeah, 30% (after this optimization) is a lot! How much complexity does it bring? Because implementing the more general solution can take a lot of time too!

Let's merge this as-is. The optimization could potentially be controversial and I think warrants some discussion.

Kerollmops:
Nice, I do this!
bors merge

saik0:
> I really thought that RoaringBitmap::max was fast enough as it is basically just fetching the last container and the last number in it. I also thought that Rust would have optimized that, but was dumb to think that, that's too complex for the optimizer to understand that by itself!

It's not O(1) for bitmaps. It must linearly scan the bitmaps in reverse order to find the highest set bit.

Kerollmops:
> It's not O(1) for bitmaps. It must linearly scan the bitmaps in reverse order to find the highest set bit.

Yeah, indeed and that can be 63 loops by call to `max`, you're right!

PR #132, opened 2022-01-10
Make append O(N)

saik0:
Continuation on #131 

## Summary

* Append is currently O(N logN) as it must binary search through containers on each insert. Inserting into an array container is also logN.
 * This converts append operation to O(N) by adding a crate public `push_unchecked` methods that should only be called once it's been verified that the value >= max. Eliminating both of the logN lookups

## Before
Screen Shot 2022-01-09 at 11 13 35 PM
## After
Screen Shot 2022-01-10 at 4 40 30 AM

### DIFF
--- a/src/bitmap/store.rs
+++ b/src/bitmap/store.rs
@@ -136,6 +136,18 @@ impl Store {
         }
     }
 
+    /// Push `index` at the end of the store.
+    /// It is up to the caller to have validated index > self.max()
+    pub(crate) fn push_unchecked(&mut self, index: u16) {
+        match self {
+            Array(vec) => vec.push(index),
+            Bitmap(bits) => {
+                let (key, bit) = (key(index), bit(index));
+                bits[key] |= 1 << bit;
+            }
+        }
+    }
+
     pub fn remove(&mut self, index: u16) -> bool {
         match *self {
             Array(ref mut vec) => vec.binary_search(&index).map(|loc| vec.remove(loc)).is_ok(),
--- a/src/treemap/inherent.rs
+++ b/src/treemap/inherent.rs
@@ -58,6 +58,30 @@ impl RoaringTreemap {
         self.map.entry(hi).or_insert_with(RoaringBitmap::new).push(lo)
     }
 
+    /// Pushes `value` in the bitmap only if it is greater than the current maximum value.
+    /// It is up to the caller to have validated index > self.max()
+    pub(crate) fn push_unchecked(&mut self, value: u64) {
+        let (hi, lo) = util::split(value);
+        // BTreeMap last_mut not stabilized see https://github.com/rust-lang/rust/issues/62924
+        match self.map.iter_mut().next_back() {
+            None => {
+                // The tree is empty
+                let mut rb = RoaringBitmap::new();
+                rb.push_unchecked(lo);
+                self.map.insert(hi, rb);
+            }
+            Some((&key, rb)) => {
+                if key == hi {
+                    rb.push_unchecked(lo);
+                } else {
+                    let mut rb = RoaringBitmap::new();
+                    rb.push_unchecked(lo);
+                    self.map.insert(hi, rb);
+                }
+            }
+        }
+    }
+
     /// Removes a value from the set. Returns `true` if the value was present in the set.
     ///
     /// # Examples
--- a/src/treemap/iter.rs
+++ b/src/treemap/iter.rs
@@ -255,13 +255,30 @@ impl RoaringTreemap {
         &mut self,
         iterator: I,
     ) -> Result<u64, NonSortedIntegers> {
-        let mut count = 0;
+        // Name shadowed to prevent accidentally referencing the param
+        let mut iterator = iterator.into_iter();
+
+        let mut prev = match (iterator.next(), self.max()) {
+            (None, _) => return Ok(0),
+            (Some(first), Some(max)) if first <= max => {
+                return Err(NonSortedIntegers { valid_until: 0 })
+            }
+            (Some(first), _) => first,
+        };
+
+        // It is now guaranteed that so long as the values are iterator are monotonically
+        // increasing they must also be the greatest in the set.
+
+        self.push_unchecked(prev);
+        let mut count = 1;
 
         for value in iterator {
-            if self.push(value) {
-                count += 1;
-            } else {
+            if value <= prev {
                 return Err(NonSortedIntegers { valid_until: count });
+            } else {
+                self.push_unchecked(value);
+                prev = value;
+                count += 1;
             }
         }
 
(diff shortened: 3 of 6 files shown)


Kerollmops:
> Mistakenly opened this as a PR, rather than a draft.

Fixed that for you 😄 (top right, just below the reviewers).

saik0:
> > Mistakenly opened this as a PR, rather than a draft.
> 
> Fixed that for you 😄 (top right, just below the reviewers).

I don't believe I can change that as a non-member

Kerollmops:
> I don't believe I can change that as a non-member

I just checked on a non-owned repo and yes you can, it is just that the link is not present as it is now a Draft, but I am pretty sure you can 😃

src/treemap/inherent.rs
```diff
@@ -58,6 +58,30 @@ impl RoaringTreemap {
         self.map.entry(hi).or_insert_with(RoaringBitmap::new).push(lo)
     }
 
+    /// Pushes `value` in the bitmap only if it is greater than the current maximum value.
+    /// It is up to the caller to have validated value > self.max()
+    pub(crate) fn push_unchecked(&mut self, value: u64) {
```
Kerollmops, line 63: Same here, `unsafe`?

src/bitmap/store.rs
```diff
@@ -136,6 +136,18 @@ impl Store {
         }
     }
 
+    /// Push `index` at the end of the store.
+    /// It is up to the caller to have validated index > self.max()
+    pub(crate) fn push_unchecked(&mut self, index: u16) {
```
Kerollmops, line 141: Same here, `unsafe`?

Kerollmops (changes requested):

saik0:
These methods are not `unsafe` their are just "dangerous" 😅

The `unsafe` keyword has a very precise meaning:
> Code or interfaces whose memory safety cannot be verified by the type system.

`push_unchecked` does not violate memory safety, though care does need to be taken when calling it, which is why i suffixed it with `unchecked`.

What would you think about adding some `debug_assert` to the bodies to check the invariants in debug builds?

saik0:
Also it's `pub(crate)` so end users can't call it. Us contributors are the only ones who need to be careful.

Kerollmops:
Ok, it seems like you are right, [for dangerous internal functions we can keep them safe](https://github.com/rust-secure-code/safety-dance/issues/53). Take my comment for #127 too!

saik0:
I've added debug assertions and clear docs about the validity guarantees.

saik0:
Re: Naming

`Bitmap8k` and `SortedU16Vec` are named after _what they are_. However, if we're being honest with ourselves they are not general purpose types.

We could also name them after _what they're for_. `BitmapStore` `ArrayStore` and possibly rename `Store` enum to `StoreKind` or `StoreType` and move all three into a `store` module.

I don't have strong opinions on this. I'm just trying to think of what would be least surprising to me if I were seeing the codebase for the first time.

src/bitmap/store.rs
```diff
@@ -136,6 +136,28 @@ impl Store {
...
     }
 
+    ///
+    /// Pushes `index` in the store only if it is greater than the current maximum value.
+    /// It is up to the caller to have validated value > self.max()
+    ///
+    /// # Panics
+    ///
+    /// If debug_assertions enabled and value is > self.max()
+    pub(crate) fn push_unchecked(&mut self, index: u16) {
```
Kerollmops, line 146: Maybe we can simplify `push` here by just calling `push_unchecked` when we know that the push can be done safely?

src/bitmap/store.rs
```diff
@@ -136,6 +136,28 @@ impl Store {
...
+        }
+        match self {
+            Array(vec) => vec.push(index),
+            Bitmap(bits) => {
+                let (key, bit) = (key(index), bit(index));
+                bits[key] |= 1 << bit;
+            }
+        }
+    }
+
```
Kerollmops, line 160:
```suggestion
        if cfg!(debug_assertions) {
            assert!(self.push(index), "container max >= index")
        } else {
            match self {
                Array(vec) => vec.push(index),
                Bitmap(bits) => {
                    let (key, bit) = (key(index), bit(index));
                    bits[key] |= 1 << bit;
                }
            }
        }
    }

```

src/treemap/iter.rs
```diff
@@ -255,13 +255,30 @@ impl RoaringTreemap {
         &mut self,
         iterator: I,
     ) -> Result<u64, NonSortedIntegers> {
-        let mut count = 0;
+        // Name shadowed to prevent accidentally referencing the param
+        let mut iterator = iterator.into_iter();
+
```
Kerollmops, line 260:
```suggestion
        let mut iterator = iterator.into_iter();
```

src/treemap/iter.rs
```diff
@@ -255,13 +255,30 @@ impl RoaringTreemap {
...
+                return Err(NonSortedIntegers { valid_until: 0 })
+            }
+            (Some(first), _) => first,
+        };
+
+        // It is now guaranteed that so long as the values are iterator are monotonically
+        // increasing they must also be the greatest in the set.
+
+        self.push_unchecked(prev);
+        let mut count = 1;
```
Kerollmops, line 274:
```suggestion

        let mut count = 1;
```

src/treemap/iter.rs
```diff
@@ -255,13 +255,30 @@ impl RoaringTreemap {
...
+        let mut prev = match (iterator.next(), self.max()) {
+            (None, _) => return Ok(0),
+            (Some(first), Some(max)) if first <= max => {
+                return Err(NonSortedIntegers { valid_until: 0 })
+            }
+            (Some(first), _) => first,
+        };
+
+        // It is now guaranteed that so long as the values are iterator are monotonically
+        // increasing they must also be the greatest in the set.
```
Kerollmops, line 270:
```suggestion
        // It is now guaranteed that so long as the values of the iterator are
        // monotonically increasing they must also be the greatest in the set.
```

Kerollmops (changes requested):

src/bitmap/store.rs
saik0: Cant do this with your other suggestion of having push call `push_unchecked`

Kerollmops: ahah yeah, obviously, sorry didn't though about both at the same time 😄

src/bitmap/store.rs
Kerollmops: Can you maybe call `push_unchecked` from `push`, then?

Kerollmops (changes requested):

Kerollmops (approved):
Thank you very much for your work!
bors merge

PR #138, opened 2022-01-12
Introduce methods that check the validity of stores when deserializing

### LINKED ISSUE #137: Feature: Validate deserialize input
# Problem

Deserialize does not validate input. This could lead to a deserialized bitmap being in an invalid state.

An array container might contain duplicates or be out of order.
A bitmap container might not have the deserialized cardinality.

# Proposal
We validate the input to ensure the serialized bitmap is valid

Kerollmops:
Fixes #137.

### DIFF
--- a/src/bitmap/serialization.rs
+++ b/src/bitmap/serialization.rs
@@ -1,6 +1,7 @@
 use bytemuck::cast_slice_mut;
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
-use std::io;
+use std::convert::TryFrom;
+use std::io::{self, Error, ErrorKind};
 
 use super::container::Container;
 use crate::bitmap::store::{ArrayStore, BitmapStore, Store};
@@ -99,8 +100,9 @@ impl RoaringBitmap {
     }
 
     /// Deserialize a bitmap into memory from [the standard Roaring on-disk
-    /// format][format].  This is compatible with the official C/C++, Java and
-    /// Go implementations.
+    /// format][format]. This is compatible with the official C/C++, Java and
+    /// Go implementations. This method checks that most of the internal values
+    /// are valid.
     ///
     /// [format]: https://github.com/RoaringBitmap/RoaringFormatSpec
     ///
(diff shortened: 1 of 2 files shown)


src/bitmap/serialization.rs
```diff
@@ -99,8 +100,9 @@ impl RoaringBitmap {
     }
 
     /// Deserialize a bitmap into memory from [the standard Roaring on-disk
-    /// format][format].  This is compatible with the official C/C++, Java and
-    /// Go implementations.
+    /// format][format]. This is compatible with the official C/C++, Java and
+    /// Go implementations. This method checks that most of the internal values
```
saik0, line 104:
```suggestion
    /// Go implementations. This method checks that all of the internal values
```

src/treemap/serialization.rs
```diff
@@ -52,7 +52,9 @@ impl RoaringTreemap {
     }
 
     /// Deserialize a bitmap into memory.
+    ///
     /// This is compatible with the official C/C++, Java and Go implementations.
+    /// This method checks that most of the internal values are valid.
```
saik0, line 57:
```suggestion
    /// This method checks that all of the internal values are valid.
```

src/bitmap/serialization.rs
```diff
@@ -116,20 +118,67 @@ impl RoaringBitmap {
...
+    /// use roaring::RoaringBitmap;
+    ///
+    /// let rb1: RoaringBitmap = (1..4).collect();
+    /// let mut bytes = vec![];
+    /// rb1.serialize_into(&mut bytes).unwrap();
+    /// let rb2 = RoaringBitmap::deserialize_unchecked_from(&bytes[..]).unwrap();
+    ///
+    /// assert_eq!(rb1, rb2);
+    /// ```
+    pub fn deserialize_unchecked_from<R: io::Read>(reader: R) -> io::Result<RoaringBitmap> {
```
saik0, line 151:
```suggestion
    pub fn deserialize_from_unchecked<R: io::Read>(reader: R) -> io::Result<RoaringBitmap> {
```

src/treemap/serialization.rs
```diff
@@ -80,4 +82,38 @@ impl RoaringTreemap {
...
+    /// use roaring::RoaringTreemap;
+    ///
+    /// let rb1: RoaringTreemap = (1..4).collect();
+    /// let mut bytes = vec![];
+    /// rb1.serialize_into(&mut bytes).unwrap();
+    /// let rb2 = RoaringTreemap::deserialize_from(&bytes[..]).unwrap();
+    ///
+    /// assert_eq!(rb1, rb2);
+    /// ```
+    pub fn deserialize_unchecked_from<R: io::Read>(mut reader: R) -> io::Result<Self> {
```
saik0, line 105:
```suggestion
    pub fn deserialize_from_unchecked<R: io::Read>(mut reader: R) -> io::Result<Self> {
```

saik0 (approved):
As discussed in #127: I'm of the opinion that we do not need to expose an unchecked version. In my benchmarks the penalty was ~1ms per MB.

Given the old behavior was not to check:
1. We are introducing a performance regression
2. New versions of the lib may error on invalid inputs that old versions would have silently read

I think it makes more sense to either give the new, safe version a different name like `try_deserialize_from`, or just accept 1 & 2 and don't provide an escape hatch unless someone gives a compelling use case (ah, the benefits of being in 0.x)

Approved, with a couple small suggestions. Because the code is good, regardless of my opinions. :)

Kerollmops:
> As discussed in #127: I'm of the opinion that we do not need to expose an unchecked version. In my benchmarks the penalty was ~1ms per MB.

Hum... you are absolutely right, I will just handle the possible errors caught when checking the deserialized stores.
Thank you for the review!

Kerollmops:
bors merge

PR #139, opened 2022-01-12
Add iter ops

### LINKED ISSUE #57: Set operations for multiple maps at a time

saik0:
Closes #58
Closes #109
Fixes #57

Adds operators to iterators of bitmaps

# TODO
 - [ ] Treemap ops
 - [ ] Tests
 - [ ]  Examples in docs

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,3 +45,38 @@ impl fmt::Display for NonSortedIntegers {
 }
 
 impl Error for NonSortedIntegers {}
+
+/// An [`IntoIterator`] blanket implementation that provides extra methods for [`RoaringBitmap`]
+/// and [`RoaringTreemap`]
+///
+/// This trait is parameterized by a sealed trait and cannot be implemented for types outside
+/// of this crate
+pub trait IterExt<T>: IntoIterator<Item = T>
+where
+    T: private::Roaring,
+{
+    /// The type of output from operations.
+    type Bitmap;
+
+    /// The `union` between all elements.
+    fn or(self) -> Self::Bitmap;
+
+    /// The `intersection` between all elements.
+    fn and(self) -> Self::Bitmap;
+
+    /// The `difference` between all elements.
+    fn sub(self) -> Self::Bitmap;
+
+    /// The `symmetric difference` between all elements.
+    fn xor(self) -> Self::Bitmap;
+}
+
+mod private {
+    use crate::{RoaringBitmap, RoaringTreemap};
+
+    pub trait Roaring {}
+    impl Roaring for RoaringBitmap {}
+    impl Roaring for &RoaringBitmap {}
+    impl Roaring for RoaringTreemap {}
+    impl Roaring for &RoaringTreemap {}
+}
--- a/benchmarks/benches/lib.rs
+++ b/benchmarks/benches/lib.rs
@@ -5,8 +5,8 @@ use std::num::ParseIntError;
 use std::path::{Path, PathBuf};
 use std::{fs, io};
 
-use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion};
-use roaring::RoaringBitmap;
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, Bencher, Criterion};
+use roaring::{IterExt, RoaringBitmap};
 
 fn create(c: &mut Criterion) {
     c.bench_function("create", |b| {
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -45,3 +45,38 @@ impl fmt::Display for NonSortedIntegers {
...
 impl Error for NonSortedIntegers {}
+
+/// An [`IntoIterator`] blanket implementation that provides extra methods for [`RoaringBitmap`]
+/// and [`RoaringTreemap`]
+///
+/// This trait is parameterized by a sealed trait and cannot be implemented for types outside
+/// of this crate
+pub trait IterExt<T>: IntoIterator<Item = T>
+where
+    T: private::Roaring,
```
Kerollmops, line 56: I am not sure why you want to restrict this trait to our library types? What's dangerous in implementing that on, for example, a wrapper of `RoaringBitmap/Treemap`?

src/lib.rs
```diff
@@ -45,3 +45,38 @@ impl fmt::Display for NonSortedIntegers {
 }
 
 impl Error for NonSortedIntegers {}
+
+/// An [`IntoIterator`] blanket implementation that provides extra methods for [`RoaringBitmap`]
+/// and [`RoaringTreemap`]
+///
+/// This trait is parameterized by a sealed trait and cannot be implemented for types outside
+/// of this crate
+pub trait IterExt<T>: IntoIterator<Item = T>
```
Kerollmops, line 54: Maybe we could find a better name for this trait, as it is a trait that provides set operations on multiple sets at the same time, maybe `MultiOps` or something like that? What do you think?

Kerollmops (changes requested):

src/lib.rs
saik0: Adding new non-defaulted methods would be a breaking change downstream. I was thinking this was for public API compatibility.

src/lib.rs
saik0: Sure. We could also create a Roaring prelude that includes iter extensions, RoaringBitmap, RoaringTreemap

saik0:
I see value in exposing an API for lazy ops regardless of whether or not we expose the iterator extensions that use them internally. To generalize to any kind of aggregation the user might need to do.

Thinking out loud in pseudo-rust about the _shape_ of the API, not the names.
```rust
struct Lazy {
  bitmap: RoaringBitmap
}

impl BitOrAssign<&Lazy> for Lazy {
 // .. do a lazy or
}

impl BitXorAssign<&Lazy> for Lazy {
 // .. do a lazy xor
}

impl RoaringBitmap {
  fn into_lazy(self) -> Lazy
}

impl Lazy {
  /// equivalent of repairAfterLazy in C, Java (Go?)
  fn into_bitmap(self) -> RoaringBitmap
}

fn main() -> {
  // Roughly equivalent to current IterExt (this does not use Cow)
  let bitmaps: Vec<RoaringBitmap> = todo!();
  bitmaps.into_iter().map(RoaringBitmap::into_lazy).reduce(|acc, other| { acc |= other; acc }).into_bitmap();

  // Now do it in parallel!
  use rayon::prelude::*;
  let more_bitmaps: Vec<RoaringBitmap> = todo!();
  bitmaps.into_par_iter().map(RoaringBitmap::into_lazy).reduce(|acc, other| { acc |= other; acc }).into_bitmap();
}
```

I'd need to think more about what it might look like for borrowed bitmaps. 🤔

src/lib.rs
Kerollmops: Hum... not sure we have enough types and traits for now for that to be interesting, but maybe later with more types.

Kerollmops:
> To generalize to any kind of aggregation the user might need to do. The intuition is to have an API that allows users to do lazy operations with a newtype, that gets "repaired" (to use the same term as other lang impls) when it's unwrapped.

I like the idea, but we must absolutely make sure to document it enough. That's not clear what a `Lazy` type does. I like the idea of wrapping the bitmaps instead of methods to do lazy operations on them. I like the `Lazy` wrapping type as it look a lot like [the `Wrapping` struct of the std](https://doc.rust-lang.org/std/num/struct.Wrapping.html).

PR #144, opened 2022-01-13
change opt-level to 2 in test profile

saik0:
I've been using this on my local for awhile to speed up tests. It will require to rebuild the universe the first test run. I'm curious how it will affect CI builds.


saik0:
If `@bors` likes this change, we should consider `opt-level = 2` for `dev` profile as well.

Kerollmops (approved):
Thank you!
bors merge

PR #148, opened 2022-01-14
Draft / Proposal. Move store logic into container

saik0:
After moving most of the store logic into `ArrayStore` and `BitmapStore` both `Container` and `Store` are _both_ delegating wrapper types.

Proposing we move the remainder of `Store`'s logic into `Container`, leaving `Store` as just a dumb enum.

For the operations: there is at least one benefit to this other than just organization:
 * For a store to change it's own variant, it must mem::swap and recursively call the operator on itself. For a container: it's just an assignment.

Notes:
 * This PR also changes the explicit `BitOr::bitor` `BitOrAssign::bitor_assign` calls to the overloaded operators. I just prefer it for readability
 * This is an example of one OP, but really we would be moving out _all_ the logic into `Container`

### DIFF
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -3,7 +3,8 @@ use std::ops::{
     BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, RangeInclusive, Sub, SubAssign,
 };
 
-use super::store::{self, Store};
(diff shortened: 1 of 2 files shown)


Kerollmops:
Yeah, that would help readability, indeed 👍

saik0:
I'm going to close this and re-evaluate after SIMD impl phase

PR #152, opened 2022-01-28
Add SIMD feature skeleton

saik0:
This PR creates an empty SIMD feature and wires it to github actions

TODO:
 - [x] Ensure test build passes CI
 - [x] Manually verify the test `test_simd_conditional_compilation` has run in CI
 - [x] Remove `test_simd_conditional_compilation` test

Verified CI in my fork. See https://github.com/saik0/roaring-rs/runs/4975294592?check_suite_focus=true#step:4:108

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,8 @@ additional terms or conditions.
 
 [real-world datasets]: https://github.com/RoaringBitmap/real-roaring-datasets
 [on the user guide]: https://bheisler.github.io/criterion.rs/book/user_guide/user_guide.html
+
+## Experimental features
+
+The `simd` feature is in active development. It has not been tested. If you would like to build with `simd` note that
+`std::simd` is only in rust nightly.
--- a/README.md
+++ b/README.md
@@ -71,4 +71,4 @@ additional terms or conditions.
 ## Experimental features
 
 The `simd` feature is in active development. It has not been tested. If you would like to build with `simd` note that
-`std::simd` is only in rust nightly.
+`std::simd` is only available in Rust nightly.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,20 +36,8 @@ jobs:
 
       - uses: actions-rs/cargo@v1
         with:
-          toolchain: nightly
-          command: test
-          args: --all-features
-
-      - uses: actions-rs/cargo@v1
-        with:
           command: test
           args: --benches --manifest-path benchmarks/Cargo.toml
-
-      - uses: actions-rs/cargo@v1
-        with:
-          toolchain: nightly
-          command: test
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -67,3 +67,8 @@ additional terms or conditions.
 
 [real-world datasets]: https://github.com/RoaringBitmap/real-roaring-datasets
 [on the user guide]: https://bheisler.github.io/criterion.rs/book/user_guide/user_guide.html
+
+## Experimental features
+
+The `simd` feature is in active development. It has not been tested. If you would like to build with `simd` note that
+`std::simd` is only in rust nightly.
```
Kerollmops, line 74:
```suggestion
`std::simd` is only available in Rust nightly.
```

Kerollmops (changes requested):
Hope one day we will be able to run test on an ARM arch!

Kerollmops:
bors try

Kerollmops (commented):
Are you OK with bors only checking stable? I would like to add nightly as possibly faillible (we don't care) WDYT?

saik0:
> Are you OK with bors only checking stable? I would like to add nightly as possibly faillible (we don't care) WDYT?

If we do that, we should pin specific nightly toolchain version, and core_simd commit, to prevent that from breaking our builds

Kerollmops (approved):
I'll merge this like that, we keep the CI like that for now.

Not sure we want to pin a nightly/commit version as I would prefer that we detect breaking changes in `std::simd`.

bors merge

saik0:
> I'll merge this like that, we keep the CI like that for now.
> 
> Not sure we want to pin a nightly/commit version as I would prefer that we detect breaking changes in `std::simd`.
> 
> bors merge

Do failing builds stop PRs from being able to merge?

saik0:
> > I'll merge this like that, we keep the CI like that for now.
> > Not sure we want to pin a nightly/commit version as I would prefer that we detect breaking changes in `std::simd`.
> > bors merge
> 
> Do failing builds stop PRs from being able to merge?



> > I'll merge this like that, we keep the CI like that for now.
> > Not sure we want to pin a nightly/commit version as I would prefer that we detect breaking changes in `std::simd`.
> > bors merge
> 
> Do failing builds stop PRs from being able to merge?

@Kerollmops pinging for visibility post merge. 🙂

Kerollmops:
I think we can setup bors to just warn us about a failing build on nightly, it implies that we must bors try before bors merge. But we could simply rely on the github CI cross/green dot as we do now!

saik0:
I was more concerned that failing build would halt PR from being merged unless they pass

Kerollmops:
I thing bors rules over (?) the github CI, it will merge even if nightly fails in the github.yml CI file.

PR #154, opened 2022-01-28
Publish MSRV + verify with CI

saik0:
This PR publishes the [Minimum Supported Rust Version](https://rust-lang.github.io/rfcs/2495-min-rust-version.html) and checks the MSRV in CI

Discovered with [`cargo msrv`](https://github.com/foresterre/cargo-msrv). For the curious, it's our dependency [retain_mut](https://crates.io/crates/retain_mut)'s usage of const_generics that makes it 1.51.0

### Motivation

User's should know what the min supported rust version of the lib is without having to build it. If we add code or a new dependency that increases the MSRV, it should require a version bump as it could be a breaking change. Therefore, we add the MSRV to our list of toolchains.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,7 @@ jobs:
           - stable
           - beta
           - nightly
+          - 1.51.0
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


Kerollmops (approved):
That's a great improvement, ty!
bors merge

PR #156, opened 2022-01-28
Reduce upper bound of keys in test bitmaps

saik0:
Generating bitmaps with containers with keys in the bounds `..=u16::MAX` resulted in many disjoint bitmaps.

This limits the upper key bounds to the number of containers.

Also sneakily increases container count to 16 🥷


Kerollmops (approved):
Nice, better tests. That kind of PR is always welcome!
bors merge

PR #157, opened 2022-01-28
Fixup: allow lint rules for old linters

saik0:
Allow for false positive lint matches that warn on MSRV builds


Kerollmops (approved):
bors merge

PR #159, opened 2022-01-28
Add array visitor

saik0:
This visitor is going to be a much more helpful abstraction with SIMD. Trying to keep my PRs incremental

### DIFF
--- a/src/bitmap/store/array_store/visitor.rs
+++ b/src/bitmap/store/array_store/visitor.rs
@@ -1,8 +1,17 @@
+/// This visitor pattern allows mutliple different algorithms to be written over the same data
+/// For example: vectorized algorithms can pass a visitor off to a scalar algorithm to finish off
+/// a tail that is not a multiple of the vector width.
+///
+/// Perhaps more importantly: it separates the set algorithms from the operations performed on
+/// their results. Future work can utilize the exiting algorithms to trivially implement
+/// computing the cardinality of an operation without materializng a new bitmap.
 pub trait BinaryOperationVisitor {
     fn visit_scalar(&mut self, value: u16);
     fn visit_slice(&mut self, values: &[u16]);
 }
 
+/// A simple visitor that stores the computation result to a Vec
+/// accessible by calling `into_inner()`
 pub struct VecWriter {
     vec: Vec<u16>,
 }
--- a/src/bitmap/store/array_store/visitor.rs
+++ b/src/bitmap/store/array_store/visitor.rs
@@ -1,0 +1,32 @@
+pub trait BinaryOperationVisitor {
+    fn visit_scalar(&mut self, value: u16);
+    fn visit_slice(&mut self, values: &[u16]);
+}
+
+pub struct VecWriter {
+    vec: Vec<u16>,
+}
+
+impl VecWriter {
+    pub fn new(capacity: usize) -> VecWriter {
+        let vec = Vec::with_capacity(capacity);
+        VecWriter { vec }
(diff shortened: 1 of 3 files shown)


src/bitmap/store/array_store/visitor.rs
```diff
@@ -0,0 +1,32 @@
+pub trait BinaryOperationVisitor {
+    fn visit_scalar(&mut self, value: u16);
+    fn visit_slice(&mut self, values: &[u16]);
```
Kerollmops, line 3: Could you maybe add a small documentation line here? Just because `visit` is not quite clear when the impl of `VecWriter` is to push in an internal Vec.

Kerollmops: In fact if it does make more sense when you'll introduce SIMD, it's maybe not that interesting to document RN.

Kerollmops (changes requested):

src/bitmap/store/array_store/visitor.rs
saik0: Out of scope for this PR but here is justification for the visitor abstraction:

```rust
pub struct CardinalityCounter {
    count: usize,
}

impl CardinalityCounter {
    pub fn new() -> CardinalityCounter {
        CardinalityCounter { count: 0 }
    }

    pub fn into_inner(self) -> usize {
        self.count
    }
}

impl ArrayBinaryOperationVisitor for CardinalityCounter {
    fn visit_vector(&mut self, _value: u16x8, mask: u8) {
        self.count += mask.count_ones() as usize;
    }

    fn visit_scalar(&mut self, _value: u16) {
        self.count += 1;
    }

    fn visit_slice(&mut self, values: &[u16]) {
        self.count += values.len();
    }
}
```

Compared to the [C implementation](https://github.com/RoaringBitmap/CRoaring/blob/master/src/array_util.c#L450)

In Rust the algos monomorphize and we get the same result without having to write everything twice.

saik0: You can pass in that visitor to any `scalar::and` `vector::and`, `scalar::or` .. and so on. This can also be used to implement Jaccard index, which might be applicable to meli?

src/bitmap/store/array_store/visitor.rs
```diff
@@ -1,8 +1,17 @@
+/// This visitor pattern allows mutliple different algorithms to be written over the same data
```
Kerollmops, line 1:
```suggestion
/// This visitor pattern allows multiple different algorithms to be written over the same data
```

Kerollmops (changes requested):
After this small change, I'll merge that!

src/bitmap/store/array_store/visitor.rs
Kerollmops: Does the Jaccard index give a number between 0 and 1 which indicates the similarity of two sets? Not sure where it can be used in milli, but maybe somewhere, I didn't check that.

Kerollmops (approved):
bors merge

src/bitmap/store/array_store/visitor.rs
saik0: @Kerollmops It could be used for any metric: Jaccard, cosine similarity, etc. Any time you'd want the cardinality of some ops without having to create or mutate a bitmap, this will be quite a bit faster

josephglanville: This is awesome.

PR #160, opened 2022-01-28
fix loop termination bug in scalar sub

saik0:
I added this regression in the first refactor that moved it out. Snuck past us both, but thankfully not the tests.


Kerollmops (approved):
Whoops, thank you!
bors merge

PR #161, opened 2022-01-28
allow lint rule to build in MSRV

saik0:
```sh
cargo +1.51.0 fmt
cargo +1.51.0 fmt --manifest-path benchmarks/Cargo.toml
cargo +1.51.0 clippy --all-targets  -- -D warnings
```

This should be the last one. 🤞 The above passes on my local machine with this change.


Kerollmops (approved):
Thank you for your hard work, merging!
bors merge

PR #162, opened 2022-01-28
add PR hook to CI

saik0:
Is there a reason we dont already do this? It would help me to know if a PR is green on CI before review process even starts.


Kerollmops (approved):
That would indeed be better, I don't know, I receive ownership of this crate 1 year ago, so...
bors merge

Issue #129, opened 2022-01-10
Sample data for benchmarking


saik0:
## Summary

The build.rs script fetches one of sample datasets (`wikileaks-noquotes_srt`) on each run. 

_This request is for _some_ mechanism be to have _all_ the datasets in `RoaringBitmap/real-roaring-datasets` reliably available for benchmarking._

## Rationale

It would be helpful to have a canonical perf baseline to ensure new work does not introduce performance regressions.

Many of our existing benchmarks use data that is trivially predicted by the branch predictor. Unfortunately this makes them worse than useless: We are benchmarking the CPU branch predictor rather than our code.

To remedy this: I've been working on a benchmark suite that uses all the sample datasets. Here's a sneak peek of a subset of the report.

![Screen Shot 2022-01-09 at 6 19 36 PM](https://user-images.githubusercontent.com/997050/148711553-82395abf-5b2b-4083-81ac-4c73606f8206.png)

In order for this to be able to be merged, we would need the datasets to be available to the benchmark suite.

## Problems with the current approach:
 * Fetches the datasets specified in build.rs on _every build_
 * Only fetches one dataset

## Potential solutions
 * Check in the unzipped data sets into the repo.
 * Check in the zip files into the repo, unpack them in build.rs
 * Include `RoaringBitmap/real-roaring-datasets` as a git submodule (leave it up to the developer to `git submodule update`) if they want to run benchmarks
 * Add some statefulness to `benchmarks/build.rs` that only fetches when necessary
   * IMO, I like this one least as it's the most work and feels alot like reimplementing `git` or `rsync`

## Considerations
 * How will any potential solution effect CI builds?
 * Could the datasets also be useful for unit testing?

Kerollmops:
That would be great indeed, making sure that all of the datasets are available to the benchmark suite. I think that my preferred attempt to do so would be the one with the _git submodule_, it seems like the easiest one to implement and to use, a simple `git submodule init/update`.

However, we _must_ absolutely make sure that the README of the benchmark crate lists the command to run before running the benchmarks. I am absolutely not comfortable enough with git submodules to be sure to not make a mistake when using them 😄 😂

Issue #45, opened 2018-01-17
len should take constant-time


Nemo157:
https://github.com/rust-lang-nursery/api-guidelines/issues/149

josephglanville:
Contant time `len()` will probably require additional storage. Atleast for my usecase that would be quite unfortunate as my program has hundreds of thousands of bitmaps at any time.


Perhaps instead the name  of `len()` to be changed to `cardinality()` or another name so it don't conflict with the API guidelines and doesn't bloat `RoaringBitmap` with extra bytes?

saik0:
#127 removes 8 bytes from each array container, so maybe we can afford a few extra bytes at the root now ;)

saik0:
This is relevant to #167. 

Cardinality of all ops can all be trivially implemented in terms of the cardinality of the intersection `|A ∩ B|`. 

```
|A \ B| = |A| − |A ∩ B|
|A ∪ B| = |A| + |B| − |A ∩ B|
|A △ B| = |A| + |B| − 2|A ∩ B|
```

Intersection cardinality is cheaper to compute, so it's both simpler to implement and has better runtime performance if we have the len cached.

PR #168, opened 2022-01-31
Fast counts

### LINKED ISSUE #167: Add fast counts
Add the ability to compute the cardinality of a set operation without materializing a new bitmap.

saik0:
See #167

Add methods to compute the cardinality of two bitmaps without materializing them. Once the naming conventions and overall structure for union is approved I will add tests and the remaining ops

TODO:
 - Bitmap
   - [x] union
   - [x] intersection
   - [x] difference
   - [x] symmetric difference
   - [x] tests
   - [x] benchmarks
     - Blocked by #129
 - Treemap
   - [x] union
   - [x] intersection
   - [x] difference
   - [x] symmetric difference
   - [x] tests


Kerollmops (approved):
Looks good to me, it would be indeed faster to compute the cardinality of operations 😃

saik0:
dd70be9 greatly simplified the implementation, though it is slower without cached len. This will be much faster once #45 is fixed

Cardinality of all ops can all be trivially implemented in terms of the cardinality of the intersection `|A ∩ B|`. 

```
|A \ B| = |A| − |A ∩ B|
|A ∪ B| = |A| + |B| − |A ∩ B|
|A △ B| = |A| + |B| − 2|A ∩ B|
```

Intersection cardinality is cheapest to compute, so it's both simpler to implement and has better runtime performance once we have the len cached.

saik0:
Benchmarks done now. It's about 0.5x compared to materializing container. This is good enough to ship even without #45 it's still a win

saik0:
Curious how much this can improve milli. I would think this would be used a lot for any document distance metrics.

saik0:
> > Curious how much this can improve milli. I would think this would be used a lot for any document distance metrics.
> 
> I'm not sure about that, the most used features of roaring are obviously set operations and multi-ops (unions of a lot of bitmaps). It can take up to 70-80% (on ~35-50ms) of the search time sometimes!
> 
> [flamegraph after.svg.zip](https://github.com/RoaringBitmap/roaring-rs/files/8048154/flamegraph.after.svg.zip)

Without seeing the use site I cant tell if you need to materialize the containers or not =P

src/treemap/ops.rs
```diff
@@ -152,6 +152,110 @@ impl RoaringTreemap {
...
+    /// This is faster and more space efficient when you're only interested in the cardinality of
+    /// the union.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use roaring::RoaringBitmap;
+    ///
+    /// let rb1: RoaringBitmap = (1..4).collect();
+    /// let rb2: RoaringBitmap = (3..5).collect();
```
Kerollmops, line 168:
```suggestion
    /// use roaring::RoaringTreemap;
    ///
    /// let rb1: RoaringTreemap = (1..4).collect();
    /// let rb2: RoaringTreemap = (3..5).collect();
```

src/treemap/ops.rs
```diff
@@ -152,6 +152,110 @@ impl RoaringTreemap {
...
+    /// ```
+    pub fn intersection_len(&self, other: &RoaringTreemap) -> u64 {
+        fn intersection_len_impl(lhs: &RoaringTreemap, rhs: &RoaringTreemap) -> u64 {
+            let mut len = 0;
+            for (key, right_rb) in &rhs.map {
+                if let Some(left_rb) = lhs.map.get(key) {
+                    len += left_rb.intersection_len(right_rb);
+                }
+            }
+            len
```
Kerollmops, line 202: Wouldn't the codegen be better if we are more explicit to Rust/LLVM on what we actually wanted to compute?

```suggestion          
            rhs.map.iter().map(|key, right_rb| {
                match lhs.map.get(key) {
                    Some(left_rb) => left_rb.intersection_len(right_rb),
                    None => 0,
                }
            })
            .sum::<u64>()
```

Kerollmops (changes requested):
Looks good to me, thank you!

src/treemap/ops.rs
saik0: Yes. I've been reading too much C lately. 😂

src/bitmap/ops.rs
```diff
@@ -155,6 +155,100 @@ impl RoaringBitmap {
...
+    ///
+    /// assert_eq!(rb1.intersection_len(&rb2), (rb1 & rb2).len());
+    /// ```
+    pub fn intersection_len(&self, other: &RoaringBitmap) -> u64 {
+        Pairs::new(&self.containers, &other.containers)
+            .map(|pair| match pair {
+                (Some(..), None) => 0,
+                (None, Some(..)) => 0,
+                (Some(lhs), Some(rhs)) => lhs.intersection_len(rhs),
+                (None, None) => unreachable!(),
```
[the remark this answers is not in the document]
saik0: I like the suggestion since if we get this invalid state adding zero is idempotent, but this is still an invalid state. I think refactoring pairs to make this illegal state unrepresentable would be ideal. Out of scope to implement here of course, but curious to hear what you think.

```rust
enum Occurence<L, R> {
    Left(L),
    Right(R),
    Both(L, R)
}

File: src/treemap/ops.rs (hunk 2 above)
Comment (saik0) [reply src/treemap/ops.rs:202]: Are we both drunk? Treemap has `pairs`

File: src/bitmap/ops.rs (hunk 3 above)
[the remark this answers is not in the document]
Comment (Kerollmops) [reply src/bitmap/ops.rs:182]: Yup, that's exactly [`EitherOrBoth` from itertools](https://docs.rs/itertools/latest/itertools/enum.EitherOrBoth.html) and I am interested in using this in the `Pairs` iterator 👍 .

File: src/treemap/ops.rs [hunk 4]
```diff
@@ -44,22 +44,14 @@ impl RoaringTreemap {
...
-            intersection_len_impl(other, self)
-        }
+        self.pairs(other)
+            .map(|pair| match pair {
+                (Some(..), None) => 0,
+                (None, Some(..)) => 0,
+                (Some(lhs), Some(rhs)) => lhs.intersection_len(rhs),
+                (None, None) => 0,
+            })
+            .sum()
```
Kerollmops, line 54: Yay, was thinking about that but forget a second after, nice change here! ❤️

Kerollmops:
> Without seeing the use site I cant tell if you need to materialize the containers or not

I assure you that those bitmaps are then intersected to reduce the number of candidates 😄
Milli doesn't only return the counts of them, however, most of the time a lot of operations are applied in unspecified orders and could probably be optimized by using some kind of lazy API.

Kerollmops (approved):
Looks good to me, thank you again!
bors merge

PR #169, opened 2022-01-31
Workaround for needless borrow lint

saik0:
Fix warning caused by differences in nightly and stable clippy lints.

This will fix the build, but I think we should be asking ourselves if we really want to checking formatting and lints with every toolchain. Fixing a warning in stable might trigger another in nightly.

### DIFF
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -222,7 +222,8 @@ impl<'a> IntoIterator for &'a Container {
     type IntoIter = Iter<'a>;
 
     fn into_iter(self) -> Iter<'a> {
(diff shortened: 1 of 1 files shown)


josephglanville:
I think it would be reasonable to just lint on one toolchain but considering how rare these issues turn out to be in practice I don't feel that strongly.

Kerollmops:
Yeah, the fact that we disallow warnings on the nightly toolchain helps us detect future Clippy issues that could happen in the future. Bors is set up to only check if the stable toolchain works this way it doesn't break when we try to merge.

Displaying a red cross when any CI has warnings is great as it helps us see it!

Kerollmops (approved):
bors merge

PR #170, opened 2022-01-31
Add select

### LINKED ISSUE #135: Feature: Add select
Add `select​(n)` which returns the nth set bit

- [x] Implementation
- #170
- #174 
- [x] Tests
- [ ] Benchmark
- Blocked by #129

saik0:
Closes #135

### DIFF
--- a/src/bitmap/store/bitmap_store.rs
+++ b/src/bitmap/store/bitmap_store.rs
@@ -224,6 +224,21 @@ impl BitmapStore {
             + (self.bits[key] << (63 - bit)).count_ones() as u64
     }
 
+    pub fn select(&self, n: u16) -> Option<u16> {
+        let mut n = n as u64;
+
+        for (key, value) in self.bits.iter().cloned().enumerate() {
+            let len = value.count_ones() as u64;
+            if n < len {
+                let index = select(value, n);
+                return Some((64 * key as u64 + index) as u16);
+            }
+            n -= len;
+        }
+
+        None
+    }
+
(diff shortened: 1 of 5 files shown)


src/bitmap/store/bitmap_store.rs
```diff
@@ -237,6 +252,16 @@ impl BitmapStore {
...
 
+// this can be done in 3 instructions on x86-64 with bmi2 with: tzcnt(pdep(1 << rank, value))
+// if n > value.count_ones() this method returns 0
+fn select(mut value: u64, n: u64) -> u64 {
+    // reset n of the least significant bits
+    for _ in 0..n {
+        value &= value - 1;
+    }
+    value.trailing_zeros() as u64
+}
```
Kerollmops, line 263: Yeah, that's indeed what should have been done! I badly understood the `select` method, I thought it was a function that was returning whether a given bit was on or off. Sorry.

src/bitmap/store/mod.rs
```diff
@@ -145,6 +145,13 @@ impl Store {
             Bitmap(bits) => bits.rank(index),
         }
     }
+
+    pub fn select(&self, n: u16) -> Option<u16> {
+        match *self {
+            Array(ref vec) => vec.select(n),
+            Bitmap(ref bits) => bits.select(n),
```
Kerollmops, line 152:
```suggestion
        match self {
            Array(vec) => vec.select(n),
            Bitmap(bits) => bits.select(n),
```

Kerollmops (changes requested):
Thank you for implementing that method!

PR #171, opened 2022-01-31
Benchmark datasets

### LINKED ISSUE #129: Sample data for benchmarking

saik0:
Closes #129

Adds CRoaring benchmark datasets. File contents are zstd compressed serialized bitmaps using a shared dictionary. All together adds about 18 MiB.

Utilizing the datasets is out of scope for this PR.


Kerollmops (changes requested):
Wasn't the initial plan to create a submodule of [the real-roaring-dataset repository](https://github.com/RoaringBitmap/real-roaring-datasets) and to let the user `submodule init/update` by himself?

saik0:
> Wasn't the initial plan to create a submodule of [the real-roaring-dataset repository](https://github.com/RoaringBitmap/real-roaring-datasets) and to let the user `submodule init/update` by himself?

Correct. This avoid a manual step.

Is there some other concern regarding adding these? Would it bloat the size of the published crate?

Kerollmops:
> Correct. This avoid a manual step.

Yup, but cloning the repository is a little issue now, 18Mo of dataset is a lot. Also, I'm not sure that we want to make git care about them. I would prefer that we depend on the official repo, even with a specific revision.

> Is there some other concern regarding adding these? Would it bloat the size of the published crate?

Yup, the other concern is about the size of the repository itself, not the crate as you put the datasets in the benchmark subcrate. I hope that cargo doesn't push useless folders!

saik0:
>  Yup, the other concern is about the size of the repository itself

The difference between this PR and all the zips is rather large  ~18 MB compared to ~95. zstd is ✨ _magic_ ✨
This size reduction was what led me to just include them in the benchmark crate.

> I hope that cargo doesn't push useless folders!

I have verified, [it does not](https://docs.rs/crate/roaring/0.8.1/source/).

My main concern is admittedly a subjective one. It just _feels icky_ 🤢  to introduce a manual fetch step to run benchmarks.

Kerollmops:
> My main concern is admittedly a subjective one. It just _feels icky_ 🤢 to introduce a manual fetch step to run benchmarks.

Maybe we can `git submodule init/update` by ourselves in the `build.rs`. I just checked the size of the repository, not even compressed, it's 628K, adding 18M to that is a lot. The other downside I see is that the user needs to install zstd to be able to run the benchmarks.

I am sure that we can find a solution to this by automating the clone in the `build.rs` script or something.  It could even just be a call to the `git submodule init/update` [`Command`](https://doc.rust-lang.org/std/process/struct.Command.html). What do you think?

saik0:
> The other downside I see is that the user needs to install zstd to be able to run the benchmarks.

This uses the zstd create under the hood, they are decompressed in the benchmark code.

> I just checked the size of the repository, not even compressed, it's 628K, adding 18M to that is a lot.

Valid.

I'm going to close this and explore using `git` binary through `Command` and also check to see if there's some other magic cargo can do. It might make sense to have a _crate_ of datasets.

Kerollmops:
> I'm going to close this and explore using `git` binary through `Command` and also check to see if there's some other magic cargo can do. It might make sense to have a _crate_ of datasets.

BTW, I don't think that creating a crate to store the benchmarks is a great idea: https://github.com/rust-lang/crates.io/issues/195.

PR #172, opened 2022-02-01
Change Cargo.toml field to "rust-version"

atouchet:
I believe this field is supposed to be "rust-version" instead of just "rust".


Kerollmops (approved):
Indeed, [you are right](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field)! I thought Cargo was throwing a warning for unknown fields.

bors merge

PR #179, opened 2022-02-07
Remove size hint and exact size from bitmap inner iterators

### LINKED ISSUE #176: Perf regression: Iteration
#141 Tracks cardinality remaining in the iterator.

The code it replaced would unconditionally panic. To guess the original author's intent: it was a runtime check that the count is tracked at the outer iterator. Panicking broke some new proptests, so we implemented the correct behavior. However, it is still true that the size hint is unreachable from the exposed API. The iterator is doing extra work the end user can never observe.

I suspect the regression went unnoticed since #125 plus this issue is still a net perf gain.

Proposing we just remove the size hints altogether and return the default `(0, None)`. It will satisfy the validity requirements of tests without needlessly slowing down iteration.

saik0:
Closes #176

### DIFF
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -241,15 +241,6 @@ impl<'a> Iterator for Iter<'a> {
     fn next(&mut self) -> Option<u32> {
         self.inner.next().map(|i| util::join(self.key, i))
     }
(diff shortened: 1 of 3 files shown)


Kerollmops (approved):
Thank you!
bors merge

PR #180, opened 2022-02-07
add unvalidated deserialization

### LINKED ISSUE #177: Non-validating deserialize
I changed my mind while preparing for release. I still think the default should validate inputs, but we should offer an escape hatch that does not validate for deserialize heavy workloads from trusted sources.

saik0:
Closes #177

### DIFF
--- a/src/bitmap/serialization.rs
+++ b/src/bitmap/serialization.rs
@@ -139,7 +139,7 @@ impl RoaringBitmap {
     /// let rb1: RoaringBitmap = (1..4).collect();
     /// let mut bytes = vec![];
     /// rb1.serialize_into(&mut bytes).unwrap();
-    /// let rb2 = RoaringBitmap::deserialize_from(&bytes[..]).unwrap();
+    /// let rb2 = RoaringBitmap::deserialize_from_unvalidated(&bytes[..]).unwrap();
     ///
     /// assert_eq!(rb1, rb2);
     /// ```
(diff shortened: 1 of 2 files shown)


src/bitmap/serialization.rs
```diff
@@ -118,7 +120,49 @@ impl RoaringBitmap {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use roaring::RoaringBitmap;
+    ///
+    /// let rb1: RoaringBitmap = (1..4).collect();
+    /// let mut bytes = vec![];
+    /// rb1.serialize_into(&mut bytes).unwrap();
+    /// let rb2 = RoaringBitmap::deserialize_from(&bytes[..]).unwrap();
```
Kerollmops, line 142:
```suggestion
    /// let rb2 = RoaringBitmap::deserialize_from_unvalidated(&bytes[..]).unwrap();
```

Kerollmops (changes requested):
Thank you, I'll merge once this little fix is done.

src/bitmap/serialization.rs
saik0: Good catch.

Kerollmops (changes requested):
We can probably add this method to the `Treemap` type, too?

saik0:
> We can probably add this method to the `Treemap` type, too?

Done.

PR #181, opened 2022-02-07
prefer flatten over flatmap identity

saik0:
Small readability improvement


Kerollmops (approved):
Thank you, that's indeed better!
bors merge

PR #182, opened 2022-02-07
Constant time len

### LINKED ISSUE #45: len should take constant-time

saik0:
Fixes #45
Closes #105

### DIFF
--- a/src/bitmap/inherent.rs
+++ b/src/bitmap/inherent.rs
@@ -15,7 +15,7 @@ impl RoaringBitmap {
     /// let mut rb = RoaringBitmap::new();
     /// ```
     pub fn new() -> RoaringBitmap {
-        RoaringBitmap { containers: Vec::new() }
+        RoaringBitmap { len: 0, containers: Vec::new() }
     }
 
     /// Adds a value to the set.
@@ -41,7 +41,9 @@ impl RoaringBitmap {
                 &mut self.containers[loc]
             }
         };
-        container.insert(index)
+        let inserted = container.insert(index);
+        self.len += inserted as u64;
+        inserted
     }
 
     /// Search for the specific container by the given key.
@@ -90,7 +92,9 @@ impl RoaringBitmap {
         // If the end range value is in the same container, just call into
         // the one container.
         if start_container_key == end_container_key {
-            return self.containers[first_index].insert_range(start_index..=end_index);
+            let inserted = self.containers[first_index].insert_range(start_index..=end_index);
+            self.len += inserted;
+            return inserted;
         }
 
         // For the first container, insert start_index..=u16::MAX, with
@@ -115,7 +119,7 @@ impl RoaringBitmap {
         let last_index = self.find_container_by_key(end_container_key);
 
         inserted += self.containers[last_index].insert_range(0..=end_index);
-
+        self.len += inserted;
         inserted
     }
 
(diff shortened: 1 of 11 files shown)


src/bitmap/iter.rs
```diff
@@ -17,16 +17,16 @@ pub struct IntoIter {
 }
 
 impl Iter<'_> {
-    fn new(containers: &[Container]) -> Iter {
-        let size_hint = containers.iter().map(|c| c.len()).sum();
-        Iter { inner: containers.iter().flatten(), size_hint }
+    fn new(bitmap: &RoaringBitmap) -> Iter {
+        let size_hint = bitmap.len();
+        Iter { inner: bitmap.containers.iter().flatten(), size_hint }
```
Kerollmops, line 22: Can you explain to me the fact #179 removes the size_hint implementation and here we continue to track the length of the bitmap? I feel lost.

src/treemap/inherent.rs
```diff
@@ -55,7 +57,9 @@ impl RoaringTreemap {
     /// ```
     pub fn push(&mut self, value: u64) -> bool {
         let (hi, lo) = util::split(value);
-        self.map.entry(hi).or_insert_with(RoaringBitmap::new).push(lo)
+        let pushed = self.map.entry(hi).or_insert_with(RoaringBitmap::new).push(lo);
+        self.len += pushed as u64;
+        pushed
```
Kerollmops, line 62: > Pushes `value` in the treemap only if it is greater than the current maximum value.

This is wrong BTW 😄

Kerollmops (changes requested):
Maybe we can use the constant `len` in select too?

src/treemap/inherent.rs
saik0: You're right! Please open an issue.

Kerollmops (changes requested):
Maybe we can use the new constant `len` in select too?

saik0:
I'm not sure what the right approach to testing this is.

Kerollmops:
> I'm not sure what the right approach to testing this is.

Do you mean benchmarking the constant length or checking that `select` which uses `len()` is valid?
You already added tests for the `select` method so... Nothing more to do.

saik0:
> However, checking that the length is valid is maybe a little bit complex

This is what I meant. I'm trying to think of a clean way to `debug_assert` that each time the len changes we compare it against the computed len

Kerollmops:
> This is what I meant. I'm trying to think of a clean way to `debug_assert` that each time the len changes we compare it against the computed len

Wouldn't it be ok to directly do `debug_assert_eq!(self.iter().count(), self.len())`? As the macro expansion is done at compile-time, therefore the call to iter/count is only done for debug builds.

I just hope that we haven't implemented the `count` method on our iterator to only call `self.len()` 😂

Kerollmops:
I just checked in the playground and the macro expansion indeed work like that: [it didn't panic when compiling in release](https://play.rust-lang.org/?version=stable&mode=release&edition=2021).

Kerollmops:
> Of course it does. Why should it recompute the cached len?

Ok so we should trick by doing a `fold` instead.

saik0:
> > Of course it does. Why should it recompute the cached len?
> 
> Ok so we should trick by doing a `fold` instead.

We'd probably have to wrap it in a `black_box`. LLVM is very clever.

src/bitmap/inherent.rs
```diff
@@ -375,8 +386,6 @@ impl RoaringBitmap {
     /// assert_eq!(rb.rank(10), 2)
     /// ```
     pub fn rank(&self, value: u32) -> u64 {
-        // if len becomes cached for RoaringBitmap: return len if len > value
```
saik0, line 378: This is incorrect. Should have been return len if len > max

Computing max is not constant time for `BitmapStore`

PR #184, opened 2022-02-07
Add into_rev_iter

schmidek:
Reverse iteration

### DIFF
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -19,6 +19,11 @@ pub struct Iter<'a> {
     inner: store::Iter<'a>,
 }
 
+pub struct RevIter {
+    pub key: u16,
+    inner: store::RevIter,
+}
+
 impl Container {
     pub fn new(key: u16) -> Container {
         Container { key, store: Store::new() }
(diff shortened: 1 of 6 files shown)


saik0:
First, Hello @schmidek. Thanks for your contribution! I'd wondered before if anyone had a use case for reverse iteration. If you enjoy a hack-and-tell I'd be interested in hearing about it.

Second, Iterators have had quite a lot of code churn recently (this is my fault, sorry!). See #176 for context. I will comment on the related portions of this PR.

Last, and most relevant to these proposed changes: Reverse iteration in rust is typically implemented by calling [`rev`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rev). To enable this we would need to implement [`DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) for all our iterators. We would need a compelling reason to deviate from this expectation.

schmidek:
My use case for reverse iteration is to implement top_k both ascending and descending. 

The downside to implementing `DoubleEndedIterator` for all iterators in this case is the increased memory usage, in a lot of iterators you can implement it with no extra fields but in the `bitmap_store` iterator case I believe we'd need to essentially have an extra copy of all the fields in `BitmapIter`. Wasn't sure it was worth the trade-off for something rarely used.

saik0:
> My use case for reverse iteration is to implement top_k both ascending and descending.
> 
> The downside to implementing `DoubleEndedIterator` for all iterators in this case is the increased memory usage, in a lot of iterators you can implement it with no extra fields but in the `bitmap_store` iterator case I believe we'd need to essentially have an extra copy of all the fields in `BitmapIter`. Wasn't sure it was worth the trade-off for something rarely used.

Fair, but if I'm not mistaken in most case llvm will optimize the iterator away to an unrolled loop and the unused variables will fall out when tree-shaking.

schmidek:
Good point, I'm happy to contribute a DoubleEndedIterator implementation instead

schmidek:
Closing in favor of https://github.com/RoaringBitmap/roaring-rs/pull/187

PR #185, opened 2022-02-08
Add a submodule of a set of real datasets for benchmarks

Kerollmops:
This PR brings a set of datasets into the repository without impacting the size of the current repository.

It modifies the `build.rs` file to call `git submodule update --init` if the submodule has been downloaded before and automatically uncompress the zipped files into a folder, one folder by dataset. This is the continuation of #171.

Some parts of this PR aren't great but I have no such idea on how to fix them:
 - The datasets paths are relative to where you execute the benchmarks in the `datasets_paths.rs` module.
 - When the `build.rs` uncompress the files into subfolders inside of the `real-roaring-datasets` submodule, git shows that there are untracked changes in this folder. Should we decompress them in the `$OUT` cargo folder instead?
 - A `cargo clean` should probably clear the `real-roaring-datasets` extracted subfolders? Maybe not...

What do you think about this PR @saik0?


saik0:
I think this is a good proof of concept but agree all the points you mentioned are issues.

> The datasets paths are relative to where you execute the benchmarks in the datasets_paths.rs module.

According to to cargo book:

> In general, build scripts should not modify any files outside of OUT_DIR. It may seem fine on the first blush, but it does cause problems when you use such crate as a dependency, because there's an implicit invariant that sources in .cargo/registry should be immutable. cargo won't allow such scripts when packaging.

If you squint your eyes a bit, one way to look at this is the build system equivalent of UB. It turns out that modifying files outside of `OUT_DIR` _happens to work_ today because this is not published or used as a lib. There's no guarantee some future version of cargo wont enforce this invariant, or worse: wont, but it breaks something.

> When the build.rs uncompress the files into subfolders inside of the real-roaring-datasets submodule, git shows that there are untracked changes in this folder. Should we decompress them in the $OUT cargo folder instead?

IMO: yes, see above point.

> A cargo clean should probably clear the real-roaring-datasets extracted subfolders? Maybe not...

If they are build artifacts then yes. If they are part of our repo then no.

saik0:
Secondary issue: `Command` assumes the presence of `git` and `unzip` in the build environment.

We should consider the treade-offs between this vs adding `git2` and `zip-rs` as `build-dependencies`

saik0:
Thinking out loud: Why are we doing this as a build script time instead of run time in the first place?

PR #186, opened 2022-02-09
add runtime dataset fetch and parse in-place

### LINKED ISSUE #129: Sample data for benchmarking

saik0:
Closes #129
Closes #171
Closes #185

Here's my go at fetching the datasets at runtime

 * Datasets are lazily fetched the first time they're needed (or updated, if local `HEAD != origin/master`).
 * The zip files are parsed-in place on every benchmark run, to keep the on-disk size down.
 * The parsing is also lazy, and happens at most once.
 * This PR updates any benchmarks that were already using limited data from `wikileaks-noquotes` to use all the datasets.
 * A fast follow PR will update all the benchmarks.

@Kerollmops Third times the charm?


benchmarks/Cargo.toml
```diff
@@ -10,17 +10,14 @@ publish = false
 roaring = { path = ".." }
 
 [dev-dependencies]
+once_cell = "1.9.0"
+git2 = "0.13.25"
```
Kerollmops: This is frustrating that I can't compile due to the fact that OpenSSL isn't reachable, I prefer that we use this feature. This is also the reason why I prefer using `Command` with `git` and `unzip` as the benchmarks are used by developers and having `git` and `un/zip` is quite common for developpers.

```suggestion
git2 = { version = "0.13.25", features = ["vendored-openssl"] }
```

benchmarks/benches/lib.rs
```diff
@@ -1,14 +1,16 @@
-mod datasets_paths;
-
 use std::cmp::Reverse;
-use std::convert::TryInto;
 use std::num::ParseIntError;
 use std::path::{Path, PathBuf};
 use std::{fs, io};
```
Kerollmops:
```suggestion
```

Kerollmops (changes requested):
Thank you very much, that is quite good! Even the little `indicatif` progress bar ❤️

benchmarks/benches/datasets.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    Ok(repo_path)
+}
+
+fn parse_datasets<P: AsRef<Path>>(path: P) -> Result<Vec<Dataset>, Box<dyn std::error::Error>> {
+    const DATASET_FILENAME_WHITELIST: &[&str] = &[
+        "census-income.zip",
+        "census-income_srt.zip",
+        "census1881.zip",
+        "census1881_srt.zip",
```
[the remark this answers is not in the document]
saik0: They weren't in the paper. You may have noticed, but I've been a a _slightly_ singularly focused.

benchmarks/benches/datasets.rs
```diff
@@ -0,0 +1,186 @@
...
+
+                pb.set_position(pb.position() + size);
+            }
+
+            pb.finish();
+            datasets.push(Dataset { name, bitmaps });
+        }
+    }
+    datasets.sort_unstable_by(|a, b| a.name.cmp(&b.name));
+    println!();
```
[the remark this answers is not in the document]
saik0: Yes, after the progress bars, a little breathing room for criterion's output.

benchmarks/benches/datasets.rs
```diff
@@ -0,0 +1,186 @@
...
+pub struct Dataset {
+    pub name: String,
+    pub bitmaps: Vec<RoaringBitmap>,
+}
+
+fn init_datasets() -> Result<PathBuf, Box<dyn std::error::Error>> {
+    let out_dir = env::var_os("CARGO_MANIFEST_DIR").ok_or(env::VarError::NotPresent)?;
+
+    let out_path = Path::new(&out_dir);
+    let repo_path = out_path.join("real-roaring-datasets");
```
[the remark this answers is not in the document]
saik0: Correct.

benchmarks/benches/datasets.rs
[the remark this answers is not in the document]
Kerollmops: As you which, I don't know why those datasets are in this repo, thought.

benchmarks/Cargo.toml
saik0: Looking at git2 `Cargo.toml`

```toml
[features]
unstable = []
default = ["ssh", "https", "ssh_key_from_memory"]
ssh = ["libgit2-sys/ssh"]
https = ["libgit2-sys/https", "openssl-sys", "openssl-probe"]
vendored-libgit2 = ["libgit2-sys/vendored"]
vendored-openssl = ["openssl-sys/vendored", "libgit2-sys/vendored-openssl"]
ssh_key_from_memory = ["libgit2-sys/ssh_key_from_memory"]
zlib-ng-compat = ["libgit2-sys/zlib-ng-compat"]
```

I've disabled all features now, so If I'm not mistaken: no default `https`, so no transitive `openssl-sys`

Kerollmops: I just tried with `git2 = { version = "0.13.25", default-features = false }` and indeed it works great!

saik0:
CI is failing in my branch from git. Investigating.

.github/workflows/test.yml
```diff
@@ -19,66 +19,87 @@ jobs:
           - beta
           - nightly
           - 1.56.1
+    env:
+      RUSTFLAGS: "-C target-cpu=native -C opt-level=3"
```
saik0, line 23: MSRV cargo does not know about profiles, so it was running with opt-level 0.

This is going to make out CI much faster :)

benchmarks/benches/datasets.rs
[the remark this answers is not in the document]
saik0: I cant say I know why _any_ of these datasets are chosen. I'd need some kind of data vis to see them. The only one I've gotten familiar with is weather_sept_85. It's sparse so I used it often to test array-array ops.

.github/workflows/test.yml
```diff
@@ -19,66 +19,87 @@ jobs:
...
           command: build
+          args: --all-targets
 
-      - uses: actions-rs/cargo@v1
+      - name: Build benchmarks
+        uses: actions-rs/cargo@v1
         with:
-          command: test
+          command: build
+          args: --manifest-path benchmarks/Cargo.toml --all-targets
```
saik0, line 54: I think with a workspace we wont need all these duplicated entries. I've called it out in #189

.github/workflows/test.yml
```diff
@@ -19,66 +19,87 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - name: Checkout roaring-rs
+        uses: actions/checkout@v2
 
-      - uses: actions-rs/toolchain@v1
+      - name: Checkout benchmark datasets
+        uses: actions/checkout@v2
+        with:
+          repository: "RoaringBitmap/real-roaring-datasets"
+          path: "benchmarks/real-roaring-datasets"
```
Kerollmops, line 34: Ok, so this makes sure that the `real-roaring-datasets` repository is cloned before we start running the benchmarks. I wasn't aware that the CI was offline, good catch!

This is probably to avoid crypto miners... I don't know.

Kerollmops (approved):
Is it ready for review? Do you think I can't take a last look at your PR?

.github/workflows/test.yml
saik0: I added vendored ssl as you suggested, I suppose windows machines might not be able to link it (?). I still prefer checking out the repo explicitly in CI. It puts the network fetch at the beginning. I've also added fetch steps before compiling begins for the same reason.

Now if theres a network failure it should always happen in the first ~30 sec or so of CI

Kerollmops (changes requested):
Your PR looks good to me!

.github/workflows/test.yml
```diff
@@ -19,66 +19,98 @@ jobs:
...
+          args: --manifest-path benchmarks/Cargo.toml --all-targets -- -D warnings
+
+      - name: Check formatting
+        uses: actions-rs/cargo@v1
         with:
           command: fmt
           args: -- --check
 
-      - uses: actions-rs/cargo@v1
+      - name: Check benchmark formatting
```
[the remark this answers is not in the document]
saik0: This is correct as is.

PR #192, opened 2022-02-10
Benchmarks using real datasets

### LINKED ISSUE #135: Feature: Add select

saik0:
Adds benchmarks using all the real datasets.

 * This benchmark suite has gotten so large that I rarely ever run the entire suite at once. I most often find myself using the regex filters like `cargo bench -- '^(:?rank|select)'` in my build-bench-optimize loop
 * My intention is to backport the benchmark crate to 0.8.1 for perf regression test.
 * It's unfortunate that criterion breaks down dimensions by `function * parameter`. The dimension I really care about is `(function + parameter) * version`
 * Throughput measurements vary wildly, as container types don't have the same performance.
 * Thinking out loud: we can always pipe the csv data into some plotter for our own analysis. The reporting isn't ideal, but this PR will allow us to collect the data.

Closes #135

### DIFF
--- a/benchmarks/benches/lib.rs
+++ b/benchmarks/benches/lib.rs
@@ -14,6 +14,7 @@ use crate::datasets::Datasets;
 
 mod datasets;
 
+#[allow(clippy::too_many_arguments)]
 fn pairwise_binary_op_matrix(
     c: &mut Criterion,
     op_name: &str,
@@ -227,7 +228,7 @@ fn and(c: &mut Criterion) {
     pairwise_binary_op_matrix(
         c,
         "and",
-        |a, b| BitAnd::bitand(a, b),
+        BitAnd::bitand,
         |a, b| BitAnd::bitand(a, b),
         |a, b| BitAnd::bitand(a, b),
         |a, b| BitAnd::bitand(a, b),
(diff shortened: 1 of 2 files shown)


saik0:
Thats a lot of benchmarks. 😶

benchmarks/benches/lib.rs
```diff
@@ -452,30 +486,163 @@ fn successive_or(c: &mut Criterion) {
     group.finish();
 }
 
+// LEGACY BENCHMARKS
+// =================
```
Kerollmops, line 490: Do we really want to keep them? Isn't this covered by other benchmarks and tests?

Kerollmops (commented):
This `pairwise_binary_op_matrix` is pretty cool, reduces the amount of code to review too!

benchmarks/benches/lib.rs
saik0: These are not helpful, but we should have _some_ meaningful way to measure `insert`, `remove`, `delete`, `insert_range` and `remove range`

saik0:
> I am worried about the amount of benchmarks criterion is doing with all those datasets

What are your concerns?

Kerollmops (approved):
Right, let's merge like that. The other concern I had was that before you told me I forgot that it was possible to filter by regex, I only remember about filtering by _contained_ strings. And that was one other reason I wanted you to document that in the benchmarks README. But we can merge this PR as is and do that later. Thank you anyway for the work you've done 🎉

bors merge

PR #199, opened 2022-02-11
rename deserialize_from_unvalidated to deserialize_unchecked_from

saik0:


Kerollmops (commented):
I am not a big fan of this name, I would prefer `nonvalidating_deserialize_from` or maybe `unchecked_deserialize_from`? What do you think?

saik0:
> I am not a big fan of this name, I would prefer `nonvalidating_deserialize_from` or maybe `unchecked_deserialize_from`? What do you think?

+1 to `unchecked_deserialize_from` but `std` convention looks to be to use a suffix so I'd prefer it.

`Option::unwrap_unchecked`
`Slice::get_unchecked`

Kerollmops (changes requested):
Using `deserialize_from_unchecked` makes the function name strange and not express what this function actually does.

I would prefer `deserialize_unchecked_from` instead. [As you can see in the std](https://doc.rust-lang.org/std/?search=unchecked) there are multiple function names containing `unchecked` word not particularly at the end of it i.e. [`Pin::get_unchecked_mut`](https://doc.rust-lang.org/std/pin/struct.Pin.html#method.get_unchecked_mut), [`i32::unchecked_add`](https://doc.rust-lang.org/std/primitive.i32.html#method.unchecked_add).

Kerollmops (approved):
Looks good to me, thank you!
bors merge

PR #202, opened 2022-02-11
specify None return cases for select

saik0:


Kerollmops (approved):
That's better indeed, thank you!
bors merge

PR #216, opened 2022-02-25
Remove the portable-simd unpublished crate

Kerollmops:
In order to be able to publish the v0.9.0 on crates.io, a crate must have all its dependencies available on crates.io. Unfortunately, the _portable-simd_ crate is not published but is part of [the _core_ library](https://doc.rust-lang.org/nightly/core/simd), we can try to use this instead.

I am encountering only one final issue with the above-defined library switch. The [`Mask<i16, 8_usize>::to_bitmask`](https://github.com/rust-lang/portable-simd/blob/78a18c3433c13bbd6259242185667c385eafe859/crates/core_simd/src/masks.rs#L228-L234) doesn't exist in the _core_ library. I must find a way to reproduce the behavior of `mask.to_bitmask()[0]` as we only need that.

The internal implementations of the `to_bitmask` method do [not look too complex](https://github.com/rust-lang/portable-simd/blob/78a18c3433c13bbd6259242185667c385eafe859/crates/core_simd/src/masks.rs#L5-L13):
 - [The generic impl](https://github.com/rust-lang/portable-simd/blob/78a18c3433c13bbd6259242185667c385eafe859/crates/core_simd/src/masks/full_masks.rs#L113-L132).
 - [The _x86_64_ impl](https://github.com/rust-lang/portable-simd/blob/78a18c3433c13bbd6259242185667c385eafe859/crates/core_simd/src/masks/bitmask.rs#L118-L124).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,11 @@ license = "MIT OR Apache-2.0"
 bytemuck = "1.7.3"
 byteorder = "1.4.3"
 retain_mut = "0.1.6"
-simd = { git = "https://github.com/rust-lang/portable-simd", package = "core_simd", optional = true }
+
+[features]
+# Enables SIMD-based algorithm that are most of the time faster than the
+# scalar ones. This feature can only be enabled on the nightly channel.
+simd = []
 
 [dev-dependencies]
 proptest = { version = "1.0.0" }
--- a/src/bitmap/store/array_store/vector.rs
+++ b/src/bitmap/store/array_store/vector.rs
@@ -11,7 +11,9 @@
 #![cfg(feature = "simd")]
 
 use super::scalar;
-use simd::{mask16x8, simd_swizzle, u16x8, LaneCount, Mask, Simd, SimdElement, SupportedLaneCount};
+use core::simd::{
+    mask16x8, simd_swizzle, u16x8, LaneCount, Mask, Simd, SimdElement, SupportedLaneCount,
+};
 
 // a one-pass SSE union algorithm
 pub fn or(lhs: &[u16], rhs: &[u16], visitor: &mut impl BinaryOperationVisitor) {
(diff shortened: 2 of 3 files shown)


saik0:
### For context

 * https://github.com/rust-lang/portable-simd/issues/204
 * https://github.com/rust-lang/portable-simd/pull/239

### In summary

`to_bitmask` is not present in `std` because it requires the `generic_const_exprs` and `std` is compiled with no features.

It's a common enough operation that supporting int/byte bitmasks without `generic_const_exprs` is being actively worked on.

### Next steps

We're fortunate in that we use `u16x8`s. The masks were interested in happen to map perfectly onto a `u8`.

I think you're suggesting we write a specialized implementation for 8 lane width bitmasks(?). Trouble is: we don't have access to `intrinsics::simd_bitmask` and would need to reimplement it. I think we'd be better off contributing to the upstream effort.

saik0:
Demo of changes we'd need to make if merged. (The git dep could be removed in favor of `std::simd`, the `Cargo.toml` changes are only to demo that there are no features)

https://github.com/saik0/roaring-rs/pull/8

Kerollmops:
Hey @saik0,

I have another idea about this issue, why can't we just release this version but we remove the `simd` feature from the _Cargo.toml_, this way no one will be able to enable it and see that it doesn't compile. We keep the `core::simd` dependency to be able to cartes.io.

I will, obviously, need to change the release notes, the parts that talk about performance improvements thanks to your SIMD-related work.

What do you think about that?

saik0:
The `ToBitmask` branch has been merged into `portable-simd/master` and is being merged into rust nightly
 * https://github.com/rust-lang/portable-simd/pull/239
 * https://github.com/rust-lang/rust/pull/94452

If there's nothing time sensitive about the release I think we should wait. I can open a PR similar to https://github.com/saik0/roaring-rs/pull/8 but based nightly once it's merged.

saik0:
Merged. See #217

PR #217, opened 2022-03-02
Replace portable-simd create with core::simd

saik0:
Closes #216


Kerollmops:
Hey,

I just tested your branch on the latest nightly channel on my MacBook Pro M1 and it was failing to compile, but don't worry it just because the latest nightly available for the _aarch64-apple-darwin_ target is from the 28th of February. I remembered [this website where you can review the availability of the different rustup targets](https://rust-lang.github.io/rustup-components-history/).

I also sent you an invitation to join this GitHub repository as a writer 🎉 I you want to join obviously!

PR #222, opened 2022-06-26
Introduce `insert_range` for RoaringTreemap

not-jan:
Implemented a somewhat naive version of `insert_range` for RoaringTreemap

~~I wanted to add a separate benchmark to compare the performance of `insert_range` with collecting from an iterator but was unable to do so because `cargo bench` would fail with a connection error every time I ran it.~~
**Edit:** This is fixed now with https://github.com/RoaringBitmap/roaring-rs/pull/225

### DIFF
--- a/benchmarks/benches/lib.rs
+++ b/benchmarks/benches/lib.rs
@@ -675,7 +675,7 @@ fn insert_range_bitmap(c: &mut Criterion) {
 }
 
 fn insert_range_treemap(c: &mut Criterion) {
-    for &size in &[10, 100, 1_000, 5_000, 10_000, 20_000] {
+    for &size in &[1_000_u64, 10_000u64, 2 * (u32::MAX as u64)] {
         let mut group = c.benchmark_group("insert_range_treemap");
         group.throughput(criterion::Throughput::Elements(size as u64));
         group.bench_function(format!("from_empty_{}", size), |b| {
--- a/benchmarks/benches/lib.rs
+++ b/benchmarks/benches/lib.rs
@@ -8,7 +8,7 @@ use criterion::{
     Throughput,
 };
 
-use roaring::{MultiOps, RoaringBitmap};
+use roaring::{MultiOps, RoaringBitmap, RoaringTreemap};
 
 use crate::datasets::Datasets;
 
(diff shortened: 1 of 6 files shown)


Kerollmops:
Hey @not-jan,

Thank you very much for your work, I appreciate the `RoaringTreemap::insert_range` method, however, I would prefer that we don't remove the `retain_mut` dependency right now as I described in #224. Would it be possible to remove this commit from your PR (and description), please?

About the benchmark issue, I think GitHub broke URLs like _git://github.com/orga/repo.git_. I don't why it was working before but I just made #225 that you can rebase on. Sorry for the inconvenience. If it works correctly you should be able to add a benchmark 😃

not-jan:
Thank you for the quick response :) 

I've performed the requested changes and added my benchmark. I'm not sure whether my benchmark is that effective, since I've copied it from the implementation for the RoaringBitmap. Any opinions about that?

benchmarks/benches/lib.rs
```diff
@@ -650,6 +651,30 @@ fn insert_range_bitmap(c: &mut Criterion) {
     }
 }
 
+fn insert_range_treemap(c: &mut Criterion) {
+    for &size in &[10, 100, 1_000, 5_000, 10_000, 20_000] {
```
Kerollmops, line 655: Could we add bigger values here?

benchmarks/benches/lib.rs
```diff
@@ -650,6 +651,30 @@ fn insert_range_bitmap(c: &mut Criterion) {
...
+            let bm = RoaringTreemap::new();
+            b.iter_batched(
+                || bm.clone(),
+                |mut bm| black_box(bm.insert_range(0..size)),
+                criterion::BatchSize::SmallInput,
+            )
+        });
+        group.bench_function(format!("pre_populated_{}", size), |b| {
+            let mut bm = RoaringTreemap::new();
+            bm.insert_range(0..size);
```
Kerollmops, line 668: Also we always start at zero when we call `insert_range`, we should either use bigger number or make sure we are on the boundary of two bitmaps. Because, here, we are benchmarking only the `RoaringBitmap::insert_range`.

But not sure we can do much better!

Kerollmops (changes requested):
Thank you for the changes!

benchmarks/benches/lib.rs
not-jan: I've added 2 * (2^32 -1) now to ensure that we're crossing a bitmap boundary. Is there any point in adding an additional benchmark that compares the new `insert_range` to `(0..size).collect<RoaringTreemap>()`? Gut feeling says it's gonna be awfully slow with the current implementation.

Kerollmops (approved):
Hey @not-jan,
Thank you for the time you put into this PR, I would like to do a release of roaring-rs soon. I will merge your work as is and we could improve it in future PRs.

I'll try to merge it with bors but there are merge conflicts so we'll see. If it doesn't pass, could you please rebase on main (yes main, I changed the name), please?

bors merge

not-jan:
Should be all done :)

Kerollmops (changes requested):
Thank you very much! Could you please remove the _.idea_ folder?

not-jan:
> Thank you very much! Could you please remove the _.idea_ folder?

I think I erased it from the commit history succesfully but I seem to have accidentally included 9d9b8074de7892ab4e1418e6d4db14eeea3b0819 in my rebase. I can't seem to undo that though as my reflog refuses to return to a state before the rebase started :/

Kerollmops (approved):
I'll merge into another branch and try to fix that 😃 
bors merge

Issue #224, opened 2022-06-28
Remove the dependency to the `retain_mut` crate


Kerollmops:
We are depending on [the `retain_mut` crate](https://github.com/upsuper/retain_mut) which has been deprecated on May 20 2022 as the standard library integrated this function since 1.61. However, we don't want to remove this dependency right now as it would force us to bump the MSRV to 1.61, we were in 1.56.1.

I would like to bump the MSRV to 1.61 but not right now, maybe in 6 months.

/remind me that must remove the retain_mut dependency in 6 months

Dr-Emann:
Dropped in #259

PR #225, opened 2022-06-28
Fix the benchmarks datasets URL format

Kerollmops:
We were using an URL of the form _git://github.com/orga/repo.git_ which is wrong and GitHub broke it recently. I just changed it to make it valid by using the https://github.com/orga/repo.git format.

This PR also changes the version of `retain_mut` to _=0.1.7_ to avoid CI-breaking warnings. If you want to understand the reason why I am doing that you can read more in #224.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
+/proptest-regressions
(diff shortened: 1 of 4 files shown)


Kerollmops:
Thank you very much for the review!
bors merge

Issue #228, opened 2022-07-26
no method named `lanes_eq` found for struct `Simd` in the current scope


tonyabracadabra:
```rust
error[E0599]: no method named `lanes_eq` found for struct `Simd` in the current scope
   --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/roaring-0.9.0/src/bitmap/store/array_store/vector.rs:444:13
    |
444 |         | a.lanes_eq(b.rotate_lanes_left::<7>())
    |             ^^^^^^^^ help: there is an associated function with a similar name: `lanes`
```

seems that the `SIMD` feature didn't pass the latest nightly build

PR #229, opened 2022-07-26
chore(simd): bumped to latest nightly

tonyabracadabra:
closes https://github.com/RoaringBitmap/roaring-rs/issues/228

### DIFF
--- a/src/bitmap/store/array_store/vector.rs
+++ b/src/bitmap/store/array_store/vector.rs
@@ -13,7 +13,7 @@
 use super::scalar;
 use core::simd::{
     mask16x8, simd_swizzle, u16x8, LaneCount, Mask, Simd, SimdElement, SupportedLaneCount,
-    ToBitMask,
+    ToBitMask, SimdPartialEq, SimdPartialOrd,
 };
 
 // a one-pass SSE union algorithm
@@ -36,7 +36,7 @@ pub fn or(lhs: &[u16], rhs: &[u16], visitor: &mut impl BinaryOperationVisitor) {
     #[inline]
     fn handle_vector(old: u16x8, new: u16x8, f: impl FnOnce(u16x8, u8)) {
         let tmp: u16x8 = Shr1::swizzle2(new, old);
-        let mask = 255 - tmp.lanes_eq(new).to_bitmask();
+        let mask = 255 - tmp.simd_eq(new).to_bitmask();
         f(new, mask);
     }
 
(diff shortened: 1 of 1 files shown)


Kerollmops:
Hey @tonyabracadabra,

Thank you very much for this PR! However, it looks like this function doesn't compile because the compiler isn't able to find the right implementation for [`Simd<U, LANES>`](https://doc.rust-lang.org/nightly/std/simd/trait.SimdPartialOrd.html#impl-SimdPartialOrd-for-Simd%3Cu16%2C%20LANES%3E). This function defines that `U` must implement the `SimdElement` trait for the `simd_le` method from the [`SimdPartialOrd`](https://doc.rust-lang.org/nightly/std/simd/trait.SimdPartialOrd.html#) trait to be available, but this is not what is defined in the impl of this trait, nothing is related to `U: SimdElement`.

```rust
#[inline]
fn lanes_min<U, const LANES: usize>(lhs: Simd<U, LANES>, rhs: Simd<U, LANES>) -> Simd<U, LANES>
where
    U: SimdElement + PartialOrd,
    LaneCount<LANES>: SupportedLaneCount,
{
    lhs.simd_le(rhs).select(lhs, rhs)
}
```

```
error[E0599]: no method named `simd_le` found for struct `Simd` in the current scope
   --> src/bitmap/store/array_store/vector.rs:356:9
    |
356 |     lhs.simd_le(rhs).select(lhs, rhs)
    |         ^^^^^^^ method not found in `Simd<U, LANES>`
```

So, I am not quite sure how we can keep this `lanes_min` function generic over `U` when the `SimdPartialOrd` trait impl is not generic but implemented for fixed types i.e. `i8`, `u8`, `f32`, `usize`.

Kerollmops:
So I finally made most of the code work with the latest nightly. However, I cannot make [this part compile](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=f788ad2432607d3c5c53cd8afc730506).

workingjubilee:
Hello! Sorry for the churn. Hm, interesting...

workingjubilee:
For this, in a given `Simd<U, N>` you can add the constraint directly into the `where` bounds if you want a generic formulation:

`Simd<U, N>: SimdPartialEq,`

workingjubilee:
> So I finally made most of the code work with the latest nightly. However, I cannot make [this part compile](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=f788ad2432607d3c5c53cd8afc730506).
```rust
use core::simd::{LaneCount, Mask, MaskElement, Simd, SimdElement, SimdPartialOrd, SupportedLaneCount};

#[inline]
fn lanes_min<U, const LANES: usize>(lhs: Simd<U, LANES>, rhs: Simd<U, LANES>) -> Simd<U, LANES>
where
    U: SimdElement<Mask=U> + MaskElement,
    Simd<U, LANES>: SimdPartialOrd<Mask=Mask<U, LANES>>,
    LaneCount<LANES>: SupportedLaneCount,
{
    lhs.simd_le(rhs).select(lhs, rhs)
}
```
This compiles, but I must admit this is an incredibly gnarly formulation. I am going to look into making this simpler and more inferrable on our end so it's less absurd on your end.

Kerollmops:
Thank you very much for the time taken to help us, for your information I simplified the functions signatures in #230 and made it compile ☺️

PR #236, opened 2022-08-27
Use checked_add/sub rather than checking bounds explicitly

Dr-Emann:
Also add a few more unit tests, including uses of Bound::Excluded on the lower bound of the range.


Kerollmops (changes requested):
Thank you very much! Would it be possible to `cargo fmt`, please?

Kerollmops:
Note that I just renamed the master branch to main, you should be careful.

Kerollmops (approved):
Thank you very much!
bors merge

PR #237, opened 2022-08-29
contains_range / range_cardinality functions

### LINKED ISSUE #235: Implement contains_range / range_cardinality functions
Thanks for the all the work put in. 

For my use case I want to test if there is anything set in a given range.

I know I could use rank of upper bound and subtract rank of lower bound
or create a bit set of the range and then call intersection_len
however both these options presumably will be slower than an optimized single call.

I see the java library has methods to get these results.

Would it be possible to implement these methods?

Dr-Emann:
Fixes #235

### DIFF
--- a/src/bitmap/container.rs
+++ b/src/bitmap/container.rs
@@ -94,6 +94,14 @@ impl Container {
         self.store.contains(index)
     }
 
+    pub fn contains_range(&self, range: RangeInclusive<u16>) -> bool {
+        self.store.contains_range(range)
+    }
+
+    pub fn is_full(&self) -> bool {
+        self.store.is_full()
+    }
+
     pub fn is_disjoint(&self, other: &Self) -> bool {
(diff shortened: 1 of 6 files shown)


Kerollmops (changes requested):
Hey, thank you very much for that, it looks good to me. But would it be possible to add a _quickcheck_ test that:
 - One that creates a RoaringBitmap/Treemap with a random range, adds more numbers into it (in the front and the back of that range) and check that the original range it contained in it.
  - One that does the same but we remove some random numbers from inside this range and check that the range is not contained.

Kerollmops:
Note that I just renamed the master branch to main, you should be careful.

Dr-Emann:
Good call, found a place where I had an off-by-one in the range_cardinality function.

Kerollmops (approved):
Thank you very much for your work and those new tests.
bors merge

PR #239, opened 2022-08-31
Fix range_checks proptest

### LINKED ISSUE #238: Contains range methods has a bug
Hey @Dr-Emann,

It looks like proptest [was able to find another of by one issue](https://github.com/RoaringBitmap/roaring-rs/runs/8115192793?check_suite_focus=true#step:13:253), would you mind taking a look at it, please?

Dr-Emann:
If there were duplicate values in the extra list to remove, removing
multiple instances of the same number would _not_ decrease the
range_cardinality, since the value was already removed. Changed to
expect a set of extra values, rather than a vec

Fixes #238

### DIFF
--- a/tests/range_checks.rs
+++ b/tests/range_checks.rs
@@ -1,5 +1,5 @@
 use proptest::array::uniform2;
-use proptest::collection::vec;
+use proptest::collection::hash_set;
 use proptest::prelude::*;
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
Thank you very much for the fix!
bors merge

PR #243, opened 2022-09-02
More off-by-one errors in bitmap contains_range

### LINKED ISSUE #238: Contains range methods has a bug

Dr-Emann:
The CRoaring code uses an exclusive u32 range for the implementaiton of
a bitset container's `contains_range` function.

Had to think a bit about the best way to make it work, if you can think of a prettier way to write any of it, let me know.

Fixes #238

### DIFF
--- a/src/bitmap/store/bitmap_store.rs
+++ b/src/bitmap/store/bitmap_store.rs
@@ -192,8 +192,15 @@ impl BitmapStore {
         let (start_i, start_bit) = (key(start), bit(start));
         let (end_i, end_bit) = (key(end), bit(end));
 
+        // Create a mask to exclude the first `start_bit` bits
+        // e.g. if we start at bit index 1, this will create a mask which includes all but the bit
+        // at index 0.
         let start_mask = !((1 << start_bit) - 1);
-        let end_mask = (1 << end_bit) - 1;
+        // We want to create a mask which includes the end_bit, so we create a mask of
+        // `end_bit + 1` bits. `end_bit` will be between [0, 63], so we create a mask including
+        // between [1, 64] bits. For example, if the last bit is the 0th bit, we make a mask with
+        // only the 0th bit set (one bit).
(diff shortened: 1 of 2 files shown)


Dr-Emann:
I think I thought of a better way, by being inclusive at the bit-level only

src/bitmap/store/bitmap_store.rs
```diff
@@ -403,12 +410,12 @@ impl<B: Borrow<[u64; BITMAP_LENGTH]>> DoubleEndedIterator for BitmapIter<B> {
...
-    index as usize / 64
+    usize::from(index) / 64
 }
 
 #[inline]
-pub fn bit(index: u16) -> usize {
-    index as usize % 64
+pub fn bit(index: u16) -> u32 {
+    u32::from(index) % 64
 }
```
Kerollmops, line 419: Is it a necessary change? Could you remove that from this PR, please?

Kerollmops (changes requested):
Thank you very much for your fix, I appreciate that!
Would it be possible to squash into two commits, please? One for the impl and one for the new tests?

src/bitmap/store/bitmap_store.rs
Dr-Emann: Ah, right, not anymore, it was when using `checked_shl`, which takes a u32. Removed.

Kerollmops (approved):
Thank you very much for the fix!
bors merge

Kerollmops:
bors merge

Kerollmops:
bors merge

Kerollmops:
Same here, no obvious reasons why the CI is stopped...
bors merge

PR #244, opened 2022-09-05
Array Store: Binary search for end starting at the found start

Dr-Emann:
When we've found the start of a range, when looking for the end, only
binary search in the range starting at the start we already found.

Additonally, because the array store contains no duplicates, we can use
a regular binary search for the end value, and if it's found, add one.


Kerollmops (approved):
Thank you very much, it will speed things up again!
bors merge

Kerollmops:
bors merge

Kerollmops:
I don't understand, it looks like the CI is kind of crashing when compiling or something like that...
bors merge

PR #245, opened 2022-09-05
Fix nightly clippy finding, blocking other PRs

Dr-Emann:
See https://github.com/RoaringBitmap/roaring-rs/runs/8188621171?check_suite_focus=true for example


Kerollmops (approved):
Thank you very much!
bors merge

PR #246, opened 2022-09-07
Fix MultiOps union for Result IntoIterators

jonasspinner:
This is a fix for the `MultiOps::union` operations for `IntoIterator<Item = Result<RoaringBitmap, E>>`. It previously used xor instead of or.


Kerollmops (approved):
Wow, thank you very much. I have done a bad review on our previous PR!
bors merge

PR #248, opened 2022-10-10
Implement from for arrays

michaelmior:
This is some convenience methods to allow things like `RoaringBitmap::from([10])` and `RoaringTreemap::from([1,2,3])`.

### DIFF
--- a/src/bitmap/iter.rs
+++ b/src/bitmap/iter.rs
@@ -131,6 +131,25 @@ impl IntoIterator for RoaringBitmap {
     }
 }
 
+impl From<u32> for RoaringBitmap {
+    fn from(value: u32) -> RoaringBitmap {
+        let mut rb = RoaringBitmap::new();
+        rb.insert(value);
+        rb
+    }
+}
+
+impl<const N: usize> From<[u32; N]> for RoaringBitmap {
+    fn from(mut arr: [u32; N]) -> Self {
+        if N == 0 {
+            return RoaringBitmap::new();
+        }
+
+        arr.sort();
+        RoaringBitmap::from_sorted_iter(arr.into_iter()).unwrap()
+    }
+}
+
 impl FromIterator<u32> for RoaringBitmap {
     fn from_iter<I: IntoIterator<Item = u32>>(iterator: I) -> RoaringBitmap {
         let mut rb = RoaringBitmap::new();
--- a/src/bitmap/iter.rs
+++ b/src/bitmap/iter.rs
@@ -131,14 +131,6 @@ impl IntoIterator for RoaringBitmap {
     }
 }
 
-impl From<u32> for RoaringBitmap {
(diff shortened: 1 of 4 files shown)


src/bitmap/iter.rs
```diff
@@ -131,6 +131,25 @@ impl IntoIterator for RoaringBitmap {
...
+}
+
+impl<const N: usize> From<[u32; N]> for RoaringBitmap {
+    fn from(mut arr: [u32; N]) -> Self {
+        if N == 0 {
+            return RoaringBitmap::new();
+        }
+
+        arr.sort();
+        RoaringBitmap::from_sorted_iter(arr.into_iter()).unwrap()
```
Kerollmops, line 149: I would prefer that you don't do more than what this function should do and don't sort here.

```suggestion
        RoaringBitmap::from_iter(arr.into_iter())
```

Kerollmops (changes requested):
Thank you very much, it looks great!
Would it be possible to do another PR with the same two impl, but for `u64`, for the `RoaringTreemap`, please?

michaelmior:
@Kerollmops I made the requested change. Do you want a separate PR for `RoaringTreemap` or should I add that to this one?

Dr-Emann:
I agree with the array constructor, but is `4u32.into()` really an expected thing for a container type? Vec doesn't implement From from its element type, is that something a bitmap should?

Kerollmops:
> I agree with the array constructor, but is `4u32.into()` really an expected thing for a container type? Vec doesn't implement From from its element type, is that something a bitmap should?

I agree with you I blindly accepted the change! What I would suggest would be to either not implement anything for one single `u32` as a user can simply call `RoaringBitmap::from([42])` or maybe create a clearly defined method like `RoaringBitmap::from_one(u32)`. What do you think about that @michaelmior @Dr-Emann?

> @Kerollmops I made the requested change. Do you want a separate PR for `RoaringTreemap` or should I add that to this one?

I would like that in this PR, please. But would it be possible to rename it when changes are done? Maybe only implementing the `From<[u32]>`/`From<[u64]>` thing would be enough for now?

michaelmior:
@Kerollmops I removed the implementation for u32 and added one for u64 arrays in `RoaringTreemap`

Kerollmops (approved):
bors merge

Kerollmops (approved):
bors merge

Kerollmops:
I don't know what's happening with bors. It looks like it is trying to compile the benchmarks, which is probably using to much RAM and killing the process.

Kerollmops (approved):
bors merge

Kerollmops:
Unfortunately, I am forced to merge by hand as the CI (called by bors) doesn't let me merge!

michaelmior:
@Kerollmops Thanks for the merge!

PR #257, opened 2023-07-25
Rename the `remove_front/back` methods into `remove_smallest/biggest`

Kerollmops:
This PR renames the `remove_front` and `remove_back` method into the clearer ones: `remove_smallest` and `remove_biggest`. These methods remove the `n` smallest or biggest values from the bitmaps.


ManyTheFish (approved):
Let's go

PR #260, opened 2023-08-12
Genericize RoaringBitmap and implements Roaring32 and Roaring64

grim7reaper:
`Roaring32` is the new name of `RoaringBitmap` (which is now the core generic implementation, `RoaringBitmap<V>`).
`Roaring64` is the 64-bit version and replaces `RoaringTreemap`.


src/core/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+use self::cmp::Pairs;
+use crate::Value;
+
+pub use self::iter::IntoIter;
+pub use self::iter::Iter;
+
+/// A generic compressed bitmap using the [Roaring bitmap compression scheme](https://roaringbitmap.org/).
+#[derive(PartialEq)]
+pub struct RoaringBitmap<V: Value> {
+    containers: Vec<container::Container<V>>,
```
Dr-Emann, line 29: I worry that using a Vec for RoaringBitmap64 might not be the best idea. For RoaringBitmap32, in the worst case, we're moving around 64k containers, but for RoaringBitmap64, inserting a new container may involve quite a lot of copying, which is why the original used a BTreeMap instead.

grim7reaper: I guess that's depends on the distribution of your dataset/workload, using a `BTreeMap` isn't great either in term locality/pointer chasing.

I would say that for most of the case it should be fine, at least [the paper](https://r-libre.teluq.ca/930/1/Roaring64bits.pdf) didn't notice large difference in memory usage.

But maybe both approach should be offered?

Kerollmops (changes requested):
Thank you very much for this big PR 🤗

Would it be possible to make the `IntoIter` type public, please? I was trying your new branch and was unfortunately unable to use it. I don't know if there is a flag or something to make sure cargo warns about it...

grim7reaper:
> Would it be possible to make the `IntoIter` type public, please? I was trying your new branch and was unfortunately unable to use it. I don't know if there is a flag or something to make sure cargo warns about it...

Done in the last commit
Dunno about a flag/clippy lint either.

PR #280, opened 2024-05-29
add dedicated fold and rfold impls for iterators

### LINKED ISSUE #277: Override default `fold` and `rfold` implementations for the iterators
`fold` and `rfold` are APIs for internal iteration, which is sometimes more efficient than external iteration (`next` and `next_back`). It is always possible to implement internal iteration in terms of external, which is what the default impl does, but this implementation need not be optimal. `Flatten` (which roaring uses) has more efficient `fold` and `rfold` implementations than the default. Roaring should use the implementations from `Flatten` rather than the default implementations.

Ideally you would implement `try_fold` and `try_rfold` as well, but they cannot be implemented until the `Try` trait is stabilized (https://github.com/rust-lang/rust/issues/84277).

This might apply to other default implementations overridden by `Flatten` as well.

oskgo:
Fixes #277

### DIFF
--- a/src/treemap/iter.rs
+++ b/src/treemap/iter.rs
@@ -16,11 +16,29 @@ impl<'a> Iterator for To64Iter<'a> {
     fn next(&mut self) -> Option<u64> {
         self.inner.next().map(|n| util::join(self.hi, n))
     }
+
+    #[inline]
+    fn fold<B, F>(self, init: B, mut f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        self.inner.fold(init, move |b, lo| f(b, ((self.hi as u64) << 32) + (lo as u64)))
+    }
 }
 
 impl DoubleEndedIterator for To64Iter<'_> {
     fn next_back(&mut self) -> Option<Self::Item> {
         self.inner.next_back().map(|n| util::join(self.hi, n))
+    }
+
+    #[inline]
+    fn rfold<B, F>(self, init: B, mut f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        self.inner.rfold(init, move |b, lo| f(b, ((self.hi as u64) << 32) + (lo as u64)))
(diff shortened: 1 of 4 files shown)


oskgo:
I tried benchmarking it as well, but there were server connection issues.

src/treemap/iter.rs
```diff
@@ -16,12 +16,40 @@ impl<'a> Iterator for To64Iter<'a> {
...
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        #[inline]
+        fn f_with_hi<B>(mut f: impl FnMut(B, u64) -> B, hi: u32) -> impl FnMut(B, u32) -> B {
+            move |b, lo| f(b, ((hi as u64) << 32) + (lo as u64))
+        }
+
+        self.inner.fold(init, f_with_hi(f, self.hi))
```
Kerollmops, line 31: Is the only purpose of this to force inlining the closure? Or is it because you need to pass the `self.hi` as `move`, and then it tries to pass `self` as `move`? Because I would prefer a simpler approach:
```suggestion
        self.inner.fold(init, move |b, lo| f(b, ((self.hi as u64) << 32) + (lo as u64)))
```

Kerollmops: Maybe we can move this `f_with_hi ` function into the _treemap/util.rs_ module instead of copying/pasting it everywhere and adding a small and helpful description in the process? What do you think?

Kerollmops (changes requested):

src/treemap/iter.rs
oskgo: There wasn't much thought given to it from my end. I just did what I saw in adjacent code to `Flatten` in the stdlib.

I've tracked the inlined internal functions down to this PR: https://github.com/rust-lang/rust/pull/62429

The reasoning is making the closures less generic to reduce type length and deduplicate monomorphization. I don't think those apply here, so I'll switch to using the closures directly instead.

Kerollmops (approved):
Thank you very much for the changes!

PR #281, opened 2024-06-05
Direct intersections with serialized RoaringBitmaps

Kerollmops:
This PR is related to #263 and implements methods to do intersections directly with serialized data. This avoids deserializing everything in memory to reduce it just after which drastically reduces the amount of allocated memory and intensive intersection operations.

## To Do
- [x] Improve function documentation.
- [x] Use the containers offsets when available.
- [ ] ~Convert operation methods to use the `BitAndAssign/BitAnd` traits~ (too much work for now, let's ship)

### DIFF
--- a/src/bitmap/ops_with_serialized.rs
+++ b/src/bitmap/ops_with_serialized.rs
@@ -104,19 +104,14 @@ impl RoaringBitmap {
             let mut offsets = vec![0; size];
             reader.read_exact(cast_slice_mut(&mut offsets))?;
             offsets.iter_mut().for_each(|offset| *offset = u32::from_le(*offset));
-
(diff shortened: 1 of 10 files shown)


src/bitmap/ops_with_serialized.rs
```diff
@@ -0,0 +1,301 @@
...
+                containers.push(other_container);
+            }
+        }
+
+        Ok(RoaringBitmap { containers })
+    }
+}
+
+#[cfg(test)]
+#[cfg(feature = "std")] // We need this to serialize bitmaps
```
irevoire, line 281: The whole module/file is already gated behind this feature flag

```suggestion
```

Issue #189, opened 2022-02-09
Reorganize project to use cargo workspaces


saik0:
> A workspace is a set of packages that share the same Cargo.lock and output directory

Should be neater all around, no need to tun fmt / clippy / tests with multiple invocations in local dev or CI.

PR #285, opened 2024-08-05
Introduce cargo workspaces

christianschleifer:
This PR adddresses https://github.com/RoaringBitmap/roaring-rs/issues/189.


Cargo.toml
```diff
@@ -1,35 +1,22 @@
...
-default = ["std"]
-serde = ["dep:serde", "std"]
-simd = []
-std = ["dep:bytemuck", "dep:byteorder"]
-
-[dev-dependencies]
+bincode = "1.3.3"
+bytemuck = "1.16.1"
+byteorder = "1.5.0"
+criterion = "0.5"
```
Kerollmops, line 11:
```suggestion
criterion = "0.3"
```

It seems that the CI is not happy with the rustic version because you bumped criterion to v0.5 but you need to keep v0.3.

Kerollmops (changes requested):
Thank you very much for your work, it seems perfect, but you still need to accept my change. Once accepted, I hope the CI will pass 🤞

Cargo.toml
christianschleifer: Thanks for the review!

Good catch. I elegantly ignored the MSRV declaration. I had to downgrade one more dependency. 
`cargo +1.65.0 build --release && cargo +1.65.0 bench` run successfully now. I did not run the  benchmarks to the end, but they successfully started up. 

I have added the MSRV to the benchmarking crate as well. Is there a specific reason that the roaring-rs crate MSRV [is set to `1.65.0`](https://github.com/RoaringBitmap/roaring-rs/blob/main/Cargo.toml#L4), but the CI runs with `1.66.0`?

Kerollmops: > Is there a specific reason that the roaring-rs crate MSRV [is set to `1.65.0`](https://github.com/RoaringBitmap/roaring-rs/blob/main/Cargo.toml#L4), but the CI runs with `1.66.0`?

I don't think there is any good reason for that. We should maybe unify them at some point.

Issue #286, opened 2024-08-05
Iterate from minimum value


grelner:
I have a use case where I need to iterate all values greater than a specific number. The bitmaps I'm working with are huge, so using `.iter().skipWhile` is too slow. I've had a look at the `contains` method is implemented, and a sound approach to achieve what I need would probably be to do a binary search to find the starting container, and iterate from there, introducing a new method `RoaringBitmap.iter_from(&self, min_value: u32)`

I just wanted to check if anyone have any insights or opinions on this before I start implementing, or if there is an api to achieve this already that I have somehow missed?

Kerollmops:
Hey @grelner 👋

I am wondering if it would be better to implement a new method directly on the RoaringBitmap iterator instead of having another iterator constructor.

The method signature would be `Iter::skip_to(&self, n: u32) -> SkipTo<Self>` working exactly like the `SkipWhile` iterator adapter but being much more optimized. A call to `SkipTo::next` would make the iterator skip all the useless internal collections.

I am wondering if we want to introduce four methods on the four iterators (bitmap and treemap with the ref and owned ones) or if we want to declare a new trait 🤔

What do you think?

grelner:
Hi @Kerollmops 
Implementing it on the iterators rather than a new constructor on RoaringBitmap is a great idea! It would probably be much cleaner to implement. I've looked into the constructor way, and it would require adding support functions in the entire call chain (RoaringBitmap, Container, ArrayStore, BitmapStore). Using your suggestion we can avoid most of that.

As for the trait, why not? It may be useful for someone at some point

Kerollmops:
@grelner, do you plan to implement it and propose a PR? I will be happy to review one and you should probably start without the trait, we can introduce it later in the same PR.

Dr-Emann:
In the CRoaring bindings, I used the name [`reset_at_or_after`](https://docs.rs/croaring/latest/croaring/bitmap/struct.BitmapIterator.html#method.reset_at_or_after), though that's because the c implementation (`roaring_uint32_iterator_move_equalorlarger`) allows moving both forward or backward.

grelner:
I've run into a rust quirk I haven't encountered before in my implementation. I want to preserve the `DoubleEndedIterator` after the `skip_to` call, so that you can still both go backwards and forwards from the point you skipped to. 
[...]

PR #287, opened 2024-08-06
Implement SkipTo for iter::Iter

grelner:
This is the start of the implementation of #286 .
It adds a new trait `SkipTo` which for now is only implemented for iter::Iter.

The `SkipTo` trait contains a method `skip_to(&mut self, n:u32)` which should return an `Iterator` that yields values >= `n`.
The implementation for `iter::Iter` achieves this by doing a binary search for the correct container, and then another binary search if the container is an `ArrayStore`, or forwarding like a normal `BitmapIter` until a suitable value is found for `BitmapStore`. 

Right now `SkipTo` is implemented for `iter::Iter`, but I don't think this is semantically sound, and it should probably rather be implemented for `RoaringBitmap`. In the discussion for #286 @Kerollmops wanted a similar api to `SkipWhile`, but this does not really fit, as `SkipWhile` can perform it's operation only using an existing iterator, but this is not true for `SkipTo`, as it needs direct access to the containers to do it's skip efficiently, not only the existing iterator. A reference to the container vec has been added to `iter::Iter` only for this purpose. The `iterator` returned is fully independent, and not related to the iterator on which the `skip_to` call was made at all.

This PR in it's current state implements what I need for my project, but I would like to finish it, albeit at a slightly slower pace.


src/bitmap/iter.rs
```diff
@@ -298,3 +302,62 @@ impl RoaringBitmap {
...
     }
 }
+
+impl<'a> SkipTo for Iter<'a> {
+    type Iter = SkipToIter<'a>;
+
+    fn skip_to(self, n: u32) -> Self::Iter {
+        SkipToIter::new(&self.containers, n)
+    }
+}
```
Dr-Emann, line 312: Note: using the full unmodified `self.containers` means this might skip _backwards_.

To do this optimally, we might need to somewhat re-implement `std::iter::Flatten` so we can access the `slice::Iter` directly (from which we can get the remaining slice), and access to the current container iterator. :/

grelner: Really? self.containers is just a reference to RoaringBitmap.containers. It was added in this PR

I thought about changing iter::Iter to only contain a reference to containers, an index, and the iterator for the current container, and do the flatten part in ::next(), which I guess is what you're saying. This would also allow the SkipTo iterator to go backwards.
Still, are all these changes worth it to have a double ended iterator and cover the kinda weird case of skip_to to a point the iterator has already passed?

Kerollmops: > To do this optimally, we might need to somewhat re-implement `std::iter::Flatten` so we can access the `slice::Iter` directly (from which we can get the remaining slice), and access to the current container iterator.

I agree with what @Dr-Emann explains here. This is, to me, the best way to implement this iterator. `Flatten` is not hard to implement. We can have a special `Flatten` type with all the original `Flatten` type logic but dedicated to slices and bitmaps so that it can skip groups according to [the `split` high group](https://github.com/RoaringBitmap/roaring-rs/blob/c57c476b2fbe78ba4566973c1b1accab07b0f1fc/src/bitmap/util.rs#L3-L8) (skipping whole slices or bitmaps if the number we want to find is not in this group).

> I thought about changing iter::Iter to only contain a reference to containers, an index, and the iterator for the current container, and do the flatten part in ::next(), which I guess is what you're saying. This would also allow the SkipTo iterator to go backwards.

Let's forget about the `SkipTo` wrapper and only implement a non-lazy method like [`Iterator::advance_by` method](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.advance_by) that directly advances the iterator to the requested number. This method, `skip_to`, is only available on the `bitmap::Iter/IntoIter` types, so there is no need to think about [`std::iter::Rev` iterator adaptator](https://doc.rust-lang.org/stable/std/iter/struct.Rev.html).

grelner:
Hi sorry for the slow progress, I've been on vacation too.
I've pushed some changes and I think it's ready for review now.
This implements:
- advance_to for Iter and IntoIter, which skips containers using the value key. This uses an existing iterator for data
- Bitmap::iter_from, which is faster that advance_to, because it does a binary search to find the container to start from
This change significantly changes the iterator implementations, which now implement their own flatten functionality.

grelner:
@Kerollmops sorry this 1.66 requirement keeps tripping me up. I pushed a fix for it now. Maybe that should be added to https://github.com/RoaringBitmap/roaring-rs?tab=readme-ov-file#developing

grelner:
[bench-skip_to.txt](https://github.com/user-attachments/files/16709732/bench-skip_to.txt)
cargo bench iteration results of the branch

grelner:
@Kerollmops any chance to get this reviewed and hopefully merged?

roaring/src/bitmap/store/array_store/mod.rs
```diff
@@ -11,11 +11,10 @@ use core::ops::{BitAnd, BitAndAssign, BitOr, BitXor, RangeInclusive, Sub, SubAss
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
 
+use super::bitmap_store::{bit, key, BitmapStore, BITMAP_LENGTH};
 #[cfg(not(feature = "std"))]
 use alloc::boxed::Box;
 
-use super::bitmap_store::{bit, key, BitmapStore, BITMAP_LENGTH};
-
 #[derive(Clone, Eq, PartialEq)]
```
Kerollmops, line 18: Is it cargo fmt that changed that? Can you revert it?

roaring/tests/iter_advance_to.rs
```diff
@@ -0,0 +1,79 @@
...
+#[test]
+fn into_iter_empty() {
+    let bm = RoaringBitmap::new();
+    assert_eq!(bm.into_iter().advance_to(31337).next(), None)
+}
+
+#[test]
+fn iter_from() {
+    let bm = RoaringBitmap::from([1, 2, 3, 100000, 100001]);
+    let mut i = bm.iter_from(99999);
```
Kerollmops, line 65: What is the point of having an `iter_from` method when we can just use the `.advance_to` one on an iterator?

Kerollmops (changes requested):
Thank you for your work @grelner. That's very good, actually.

I think I am not a huge fan of having a new `AdvanceToIter` wrapper. Can you tell me why the `Iter/IntoIter/advance_to` method cannot just advance the iterator with a loop like you did? It will advance container by container until it reaches the right one.

roaring/tests/iter_advance_to.rs
grelner: @Kerollmops thanks for the review, I'll take a closer look at it tomorrow. As for this particular thing, my use case has very large bitmaps and extreme throughput requirements so I added this function because of that. It uses a binary search instead of the looping done in the iterators, which gains some speed. I will point out the difference between iter_from and iter().advance_to in the documentation

roaring/tests/iter_advance_to.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[test]
+fn advance_to_with_tail_iter() {
+    let bm = RoaringBitmap::from([1, 2, 3, 100000, 100001]);
+    let mut i = bm.iter();
+    i.next_back();
+    let mut i = i.advance_to(100000);
+    assert_eq!(i.next(), Some(100000));
+    assert_eq!(i.next(), None);
+}
```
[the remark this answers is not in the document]
Dr-Emann: I still feel like the better API would be a `&mut self` `advance_to` method, which just modifies the iterator inplace (using a binary search for what containers haven't been consumed)

roaring/tests/iter_advance_to.rs
Kerollmops: Nice and wouldn't it be possible to construct a normal iterator i.e., `Iter` or `IntoIter`, with this binary search technique instead of having a special one?

roaring/tests/iter_advance_to.rs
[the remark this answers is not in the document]
grelner: The advance_to method only uses &mut self now instead of a wrapper

roaring/tests/iter_advance_to.rs
grelner: I've implemented this in the latest commit. There is no Bitmap::iter_from anymore, and the iterators do not use a wrapper in advance_to, and all uses binary search

roaring/src/bitmap/iter.rs
```diff
@@ -294,7 +316,306 @@ impl RoaringBitmap {
...
+                        advance_iter(iter_back, n);
+                        *self.iter_front_mut() = Self::empty_inner_iter();
+                        return;
+                    }
+                }
+                self.clear_containers();
+                *self.iter_front_mut() = Self::empty_inner_iter();
+                *self.iter_back_mut() = None;
+            }
+        }
```
Dr-Emann, line 423: I'm pretty sure this isn't updating the `size_hint` for the iterator.

I really think trying to keep track of the `size_hint` for these iterators was a mistake, and would rather just remove the exact size_hint tracking.

grelner: You're completely right of course. I've fixed this in the latest commit. But I agree, maybe it would be better to remove this code. It certainly would make it cleaner.

grelner:
@Kerollmops did you change something in CI? It's failing on parts of the code I never touched

Kerollmops:
Hey @grelner 👋
It's may be due to https://github.com/RoaringBitmap/roaring-rs/pull/293, would you mind rebasing on main? 🤔

roaring/src/bitmap/store/mod.rs
```diff
@@ -497,26 +510,46 @@ impl PartialEq for Store {
     }
 }
 
-impl<'a> Iterator for Iter<'a> {
+impl Iter<'_> {
+    pub fn peek(&mut self) -> Option<u16> {
```
Dr-Emann, line 514: This can take `&self`

roaring/src/bitmap/iter.rs
```diff
@@ -294,7 +316,337 @@ impl RoaringBitmap {
...
+                while let Some(next) = iter.peek() {
+                    if next < n {
+                        iter.next();
+                        items_skipped += 1;
+                    } else {
+                        return items_skipped;
+                    }
+                }
+                items_skipped
+            }
```
Dr-Emann, line 399: Both array and bitmap container iterators should be able to do this more efficiently, we should be able to implement an `store::Iter::advance_to(&mut self, n: u16)` function that can do better than `O(n)`:

For an array container, we can binary search for `n`, and use `Iterator::nth` to efficiently skip ahead to just before the element. For a bitmap container, we should be able to just directly advance `key`, and mask the value (with some edge cases where `key` is unchanged and we need to mask the current value, or `key` becomes equal to `key_back`, and we need to steal value_back).

roaring/src/bitmap/iter.rs
```diff
@@ -294,7 +316,337 @@ impl RoaringBitmap {
...
+        fn pop_container_back(&mut self) -> Option<Self::Container>;
+
+        fn drain_containers_until(&mut self, index: usize);
+
+        fn clear_containers(&mut self);
+
+        fn find_container(&self, key: u16) -> Option<usize>;
+
+        fn iter_front_mut(&mut self) -> &mut container::Iter<'a>;
+        fn iter_back_mut(&mut self) -> &mut Option<container::Iter<'a>>;
```
Dr-Emann, line 344: We should be able to avoid these, if we use `slice::Iter<'a, Container>` and `vec::IntoIter<Container>`, rather than `&[Container]` and `VecDeque<Container>`. Both implement `AsRef<[Container]>`, `Iterator` (for `next` + `nth`), and `Iterator::Item: IntoIterator<IntoIter=container::Iter>`, so it should be possible to write a function something like:

```rust
fn advance_to_impl<'a, It>(
    n: u32,
    front_iter: &mut Option<container::Iter<'a>>,
    containers: &mut It,
    back_iter: &mut Option<container::Iter<'a>>,
) where
    It: Iterator,
    It: AsRef<[Container]>,
    It::Item: IntoIterator<IntoIter = container::Iter<'a>>,
{
  let (key, index) = util::split(n);
  // Compare the key of n to the front iter's key {LT => Do nothing, EQ => advance front to index, GT => consume front iter, continue}
  
  // Binary search for the key in containers, if found, front_iter = containers.nth(idx) and advance it, else use _ = nth(idx - 1) to consume up to where it would be.

  // If containers is now empty, check the back iter's key, and possibly advance it or consume it
}
```

Which has all the logic that could be called by both advance_to implementations.

roaring/src/bitmap/store/mod.rs
```diff
@@ -25,13 +25,18 @@ pub enum Store {
...
+enum IterInner<'a> {
     Array(slice::Iter<'a, u16>),
     Vec(vec::IntoIter<u16>),
     BitmapBorrowed(BitmapIter<&'a [u64; BITMAP_LENGTH]>),
     BitmapOwned(BitmapIter<Box<[u64; BITMAP_LENGTH]>>),
 }
 
+pub struct Iter<'a> {
+    inner: IterInner<'a>,
+    size_hint: u64,
```
Dr-Emann, line 37: Sorry you went to all the trouble of adding tracking the size hint, but with #295, this can all go back away now 😢

Dr-Emann (commented):
Similar to #290, I think I have a good idea of basically exactly what I'd want this implementation to look like, let me know if you'd rather I finish this off. I know there's been a lot of back and forth on this already (add the size hint, now remove the size hint)

grelner:
> Similar to #290, I think I have a good idea of basically exactly what I'd want this implementation to look like, let me know if you'd rather I finish this off. I know there's been a lot of back and forth on this already (add the size hint, now remove the size hint)

Hi @Dr-Emann , thanks for the feedback! I'm hilariously busy at the moment and unfortunately won't have time to work on this for at least 2-3 weeks so feel free to finish up. If not I'm happy to do it when I have some time again

PR #289, opened 2024-09-02
Implement Clone on the bitmap::Iter type

Kerollmops:


roaring/src/bitmap/statistics.rs
```diff
@@ -0,0 +1,106 @@
+use core::mem;
+
+use crate::bitmap::container::Container;
+use crate::RoaringBitmap;
+
+use super::store::Store;
+
+/// Detailed statistics on the composition of a bitmap.
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub struct Statistics {
```
Dr-Emann, line 10: Probably worth a `#[non_exhaustive]`, to be able to add fields backwards-compatibly.

roaring/src/bitmap/statistics.rs
```diff
@@ -0,0 +1,106 @@
...
+            match store {
+                Store::Array(array) => {
+                    cardinality += array.len();
+                    n_values_array_containers += array.len() as u32;
+                    n_bytes_array_containers += (array.capacity() * mem::size_of::<u32>()) as u64;
+                    n_array_containers += 1;
+                }
+                Store::Bitmap(bitmap) => {
+                    cardinality += bitmap.len();
+                    n_values_bitset_containers += bitmap.len() as u32;
```
Dr-Emann, line 82: This could theoretically overflow `n_values_bitset_containers` if you have a full bitmap. Not sure if that's a problem or not.

If we ever implement run containers, that's much more likely, since it'd take much less memory to hold a totally full bitmap of all full run containers

roaring/src/bitmap/store/array_store/mod.rs
```diff
@@ -30,6 +30,10 @@ impl ArrayStore {
         ArrayStore { vec: Vec::with_capacity(capacity) }
     }
 
+    pub fn capacity(&self) -> usize {
```
Dr-Emann, line 33: We'd at least need to document that this is the capacity of containers, not the capacity of values, or maybe have a different name?

PR #290, opened 2024-09-25
Add range-based iteration

mherzl:
This change adds range-based iteration, meaning a consecutive subsequence of a RoaringBitmap can now be iterated over efficiently.

The time complexity to iterate over a range of length k in a RoaringBitmap of size n is: log(n) + k.
The log(n) term comes from a binary search to find the first element, and the k term comes from iterating over the following k-length range.

Note: The `convert_range_to_inclusive` function was refactored using `num` traits, in order to handle `u16` as well as `u32`.

Unit and property-based tests are included to verify the correctness of range iteration over both arrays and bitmaps.

### DIFF
--- a/roaring/src/bitmap/range.rs
+++ b/roaring/src/bitmap/range.rs
@@ -1,0 +1,137 @@
+use core::ops::RangeBounds;
+use core::ops::RangeInclusive;
+
+use super::container::Container;
+use super::iter;
+use super::store;
+use super::util;
+use crate::RoaringBitmap;
+
+/// Iterator over a consecutive subsequence of a bitmap.
+/// Efficient; O( log[n] + k ),
+/// where n is the bitmap's length
+/// and k is the subsequence's length.
+pub struct RangeIter<'a> {
+    first: store::StorePartIter<'a>,
+    between: iter::Iter<'a>,
+    last: store::StorePartIter<'a>,
+    // size_hint: u64,
+}
+
+impl<'a> RangeIter<'a> {
+    pub fn new<R>(containers: &'a [Container], range: R) -> RangeIter<'a>
+    where
+        R: RangeBounds<u32>,
+    {
+        let (start, end) = match util::convert_range_to_inclusive(range) {
+            Some(range) => (*range.start(), *range.end()),
+            None => return RangeIter::empty(),
+        };
+
+        let (start_key, start_low) = util::split(start);
+        let (end_key, end_low) = util::split(end);
+
+        let s = containers.binary_search_by_key(&start_key, |c| c.key);
+        let e = containers.binary_search_by_key(&end_key, |c| c.key);
+
+        if s == e {
+            // single container
+            return match s {
+                Ok(i) => RangeIter {
+                    first: Self::container_part(&containers[i], start_low..=end_low, start_key),
+                    between: iter::Iter::empty(),
+                    last: store::StorePartIter::empty(),
+                },
+                Err(_) => RangeIter::empty(), // nothing to iterate over
+            };
+        }
+
+        // multiple containers
+        let (first, inner_start) = match s {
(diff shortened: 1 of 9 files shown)


roaring/src/bitmap/range.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Iterator over a consecutive subsequence of a bitmap.
+/// Efficient; O( log[n] + k ),
+/// where n is the bitmap's length
+/// and k is the subsequence's length.
+pub struct RangeIter<'a> {
+    first: store::StorePartIter<'a>,
+    between: iter::Iter<'a>,
+    last: store::StorePartIter<'a>,
+    // size_hint: u64,
+}
```
Dr-Emann, line 19: There seems to be quite a lot of overlap with #287 here.

Dr-Emann (commented):
@Kerollmops, do we really need to keep the size_hint in our iterators? It seems wasteful that `bitmap.iter().next()` has to compute the cardinality of the whole bitmap.

I think other than not keeping an accurate size_hint, I don't see any reason why this would need a seperate iterator type (if we implement our own Flatten like we do here and #287).

This and #287 feel like two sides of the same coin: To me, this is easily and efficiently built on top of an `advance_to` & `advance_back_to`, though again, only if we could drop the need for an accurate size_hint.

Kerollmops:
Hey @mherzl, @Dr-Emann 👋

Thank you for your PR and feedback on this. I would like to know if dropping the `size_hint` of the iterator could mix both #287 and #290 in a pretty and Rust-idiomatic interface. Also, adding the `num` trait is not very convenient. We should keep the dependency list low.

I love the interface of this PR much more than #287. What do you think @Dr-Emann? Do you feel it could be possible?

roaring/src/bitmap/store/bitmap_store.rs
```diff
@@ -403,6 +409,146 @@ impl Display for Error {
...
+impl<'a> Iterator for BlockSeqIter<'a> {
+    type Item = u16;
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.block_range.len() == 0 {
+            return None;
+        }
+        let mut current = BlockPartIter::new(self.start_key, self.block_range[0]);
+        if let c @ Some(_) = current.next() {
+            return c;
+        }
```
Dr-Emann, line 545: Won't this always return the same value from `next` if there's a bit set in the "current" block, since we always reload `self.block_range[0]`? `self` is not modified in any way if we `return` here, right?

Dr-Emann:
I think this change could re-use existing store iterators (it should be possible to construct a `bitmap::store::iter::Iter` over a range without introducing new iterator types) even if we didn't drop the size_hint requirement. If we _did_ drop the `size_hint` requirement, then I think it's possible to implement this _optimally_ in terms of an `advance_to`/`advance_back_to`, and returning the same `bitmap::iter::Iter` type from `bitmap.iter()` and `bitmap.iter_range(..)`

Kerollmops:
Hey @mherzl 👋

We just merged #295 and #292. Would you mind rebasing on the main and applying the @Dr-Emann proposal? 🙂

Have a nice end of the week 🌵

Dr-Emann:
Let me know if your plans change, I'd also be happy to pick this up, I have a pretty good idea of what I'd want the implementation to look like.

mherzl:
@Dr-Emann it looks like this binary search for finding ranged iterator bounds is implemented in #296. You don't have to wait for me to add that wrapper `iter_range` of your prior comment; feel free to include it.

Kerollmops (changes requested):
Hey @mherzl 👋

I just merged #296. I'll let you rebase on main. However, I wonder if we can get rid of the `num` crate 🤔

mherzl:
Sounds good, @Kerollmops, I opened #298, which includes `iter_range`, as well as my tests from #290.

PR #292, opened 2024-10-02
Optimize BitmapIter::next

Dr-Emann:
Also, some minor cleanup to the benchmarking.

On my m1 mac, this seems pretty much faster across the board:

```
# Note: I only ran the iteration benchmarks
$ critcmp before after -t 2
group                                         after                                   before
-----                                         -----                                   ------
iteration/into_iter/census-income             1.00     22.8±1.52ms 289.4 MElem/sec    1.08     24.7±0.09ms 266.7 MElem/sec
iteration/into_iter/census-income_srt         1.00     19.6±0.10ms 296.5 MElem/sec    1.09     21.4±0.11ms 271.0 MElem/sec
iteration/into_iter/census1881_srt            1.00  1972.2±14.02µs 329.2 MElem/sec    1.08      2.1±0.02ms 303.8 MElem/sec
iteration/into_iter/weather_sept_85           1.00     43.4±0.33ms 282.9 MElem/sec    1.09     47.4±0.49ms 259.1 MElem/sec
iteration/into_iter/weather_sept_85_srt       1.00     52.4±0.41ms 293.1 MElem/sec    1.10     57.4±0.56ms 267.5 MElem/sec
iteration/into_iter/wikileaks-noquotes_srt    1.00   886.9±26.84µs 309.7 MElem/sec    1.05   935.0±26.49µs 293.8 MElem/sec
iteration/iter rev/census1881                 1.00      2.4±0.02ms 396.6 MElem/sec    1.06      2.6±0.02ms 373.2 MElem/sec
iteration/iter rev/census1881_srt             1.00      2.1±0.04ms 314.9 MElem/sec    1.03      2.1±0.02ms 306.2 MElem/sec
iteration/iter rev/wikileaks-noquotes         1.00    668.7±6.64µs 392.7 MElem/sec    1.08    721.1±3.61µs 364.2 MElem/sec
iteration/iter/census-income                  1.00     20.4±0.10ms 322.8 MElem/sec    1.10     22.4±0.11ms 294.1 MElem/sec
iteration/iter/census-income_srt              1.00     17.7±0.19ms 329.1 MElem/sec    1.10     19.5±0.36ms 297.9 MElem/sec
iteration/iter/census1881_srt                 1.00  1905.9±15.00µs 340.6 MElem/sec    1.08      2.1±0.01ms 315.3 MElem/sec
[...]

PR #295, opened 2024-10-16
Lazily compute Iterator len

### LINKED ISSUE #294: Consider removing exact size_hint tracking in iterators
See some discussions under https://github.com/RoaringBitmap/roaring-rs/pull/290#pullrequestreview-2335686711

It's not clear that keeping track of the exact size of an iterator is super useful, and it has a definite cost: it may be surprising that `my_bitmap.iter().next()` (creating an iterator, and visiting only the first value) will visit every container in the bitmap to compute the size_hint field for the iterator (which is never used).

We could either:

1. Lazily (but exactly) compute the actual remaining size in `size_hint`: this means the price is only paid when `size_hint()` is actually called.
2. Lazily *inexactly* compute a minimal/maximal remaining size (whatever we can compute in `O(1)`). This would be a breaking change: currently `Iter` implements `ExactSizeIterator` (on 64 bit platforms). However I believe that's more in line with the expectation of what `size_hint()` is expected to act like.

I think I lean toward option 1 (probably not worth a breaking change to do 2, and it's `O(n)`, but with a tiny per-container cost, so still expected to be fast). What do you think @Kerollmops?

Dr-Emann:
Lazily compute the length of an iterator, rather than doing so eagerly at creation time.

Additionally, add more efficient implementations of `count`, `nth`, and `nth_back`.

This required somewhat manually reimplementing `std::iter::Flatten`, but this is something we'll need to do anyway for #290 / #287.

This is currently built on top of #292, (so the PR currently includes commits from that PR too)

Fixes #294

### DIFF
--- a/roaring/src/bitmap/container.rs
+++ b/roaring/src/bitmap/container.rs
@@ -300,6 +300,21 @@ impl Iterator for Iter<'_> {
     fn next(&mut self) -> Option<u32> {
         self.inner.next().map(|i| util::join(self.key, i))
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.inner.size_hint()
+    }
+
+    fn count(self) -> usize
+    where
+        Self: Sized,
+    {
+        self.inner.count()
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        self.inner.nth(n).map(|i| util::join(self.key, i))
+    }
 }
 
 impl DoubleEndedIterator for Iter<'_> {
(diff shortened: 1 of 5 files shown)


Kerollmops (changes requested):
Excellent implementation of the option 1. Also, when you said that it is `O(n)`, this `n` is the number of containers, so it's ok and even only called when calling `size_hint`, so that's pretty good.

I'm just thinking that when you call collect into a `Vec`, for example, the implementation will call `size_hint` 🤔 Anyway, it's still much better.

I let you rebase on main, flatten in a single commit and I'll merge, thank you again! Nice work 👌

Dr-Emann:
@Kerollmops Believe this is ready

Kerollmops (approved):
Wonderful addition! Thank you! Merging 🙏

PR #297, opened 2024-11-11
implement range based iteration

Dr-Emann:
Implement `bitmap.iter_range()` and `bitmap.into_iter_range()`, based on `advance_to()` and `advance_back_to()`. Closes #290.

Open questions:
- [x] ~~The equivalent to `.iter_range` for `BTreeSet` is called `.range()` (and so does #290). Is that a better name: `.range`/`.iter_range`? We also need an owning version, `into_range` vs `into_iter_range`?~~ Using the names `range`/`into_range`.
- [x] ~~`.range` on `BTreeSet` panics on start > end. (or start == end when both are excluded). Should we?~~ Decided to panic on ranges to match BTreeSet. The code is less pretty having to differentiate between validly empty ranges vs invalid ranges, but probably worth it.

### DIFF
--- a/roaring/src/bitmap/util.rs
+++ b/roaring/src/bitmap/util.rs
@@ -14,30 +14,56 @@ pub fn join(high: u16, low: u16) -> u32 {
     (u32::from(high) << 16) + u32::from(low)
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum ConvertRangeError {
+    Empty,
+    StartGreaterThanEnd,
+    StartAndEndEqualExcluded,
+}
+
 /// Convert a `RangeBounds<u32>` object to `RangeInclusive<u32>`,
-pub fn convert_range_to_inclusive<R>(range: R) -> Option<RangeInclusive<u32>>
+pub fn convert_range_to_inclusive<R>(range: R) -> Result<RangeInclusive<u32>, ConvertRangeError>
 where
     R: RangeBounds<u32>,
 {
-    let start: u32 = match range.start_bound() {
-        Bound::Included(&i) => i,
-        Bound::Excluded(&i) => i.checked_add(1)?,
-        Bound::Unbounded => 0,
-    };
-    let end: u32 = match range.end_bound() {
-        Bound::Included(&i) => i,
-        Bound::Excluded(&i) => i.checked_sub(1)?,
-        Bound::Unbounded => u32::MAX,
-    };
-    if end < start {
-        return None;
+    let start_bound = range.start_bound().cloned();
+    let end_bound = range.end_bound().cloned();
+    match (start_bound, end_bound) {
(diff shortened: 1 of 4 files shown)


Kerollmops (changes requested):
Thank you for this PR @Dr-Emann,

I see that this PR is mostly the same as #298. Can you rebase on main once #299 is merged, please?

However, you propose interesting open questions that I will try to answer here.

> The equivalent to `.iter_range` for `BTreeSet` is called `.range()` (and so does https://github.com/RoaringBitmap/roaring-rs/pull/290). Is that a better name: `.range/.iter_range`? We also need an owning version, `into_range` vs `into_iter_range`?

I indeed prefer that we use the `range` naming instead of the `iter_range`. This way, as you propose, we will be closer to the standard naming. I also think that, at some point, we will want to support extracting ranges but we will probably name this method `extract_range`.

> `.range` on `BTreeSet` panics on `start > end`. (or `start == end` when both are excluded). Should we?

Yes, I think a range with a `start > end` is not expected and most likely an error. We should panic. Same for the `start == end` case.

Dr-Emann:
Hmm, the current utility to to an inclusive range can't currently differentiate `1..1` (an empty range I assume we want to allow) from `1..=0` (an empty range I assume we want to disallow.

Kerollmops:
> the current utility to to an inclusive range can't currently differentiate `1..1` (an empty range I assume we want to allow) from `1..=0` (an empty range I assume we want to disallow.

Can we do it by hand then? If we cannot then, I think, it's better to just accept them as we don't want to panic for valid empty ranges.

Dr-Emann:
@Kerollmops Looks like another dependency update broke the MSRV.

Kerollmops:
Hey @Dr-Emann 👋

> Looks like another dependency update broke the MSRV.

Would you mind rebasing on main when #289 is merged? I bumped the MSRV to 1.71.1 (again).

Kerollmops (changes requested):
Whoops, still some formating issue 😞

roaring/src/bitmap/util.rs
```diff
@@ -14,30 +14,56 @@ pub fn join(high: u16, low: u16) -> u32 {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum ConvertRangeError {
+    Empty,
+    StartGreaterThanEnd,
+    StartAndEndEqualExcluded,
+}
+
 /// Convert a `RangeBounds<u32>` object to `RangeInclusive<u32>`,
-pub fn convert_range_to_inclusive<R>(range: R) -> Option<RangeInclusive<u32>>
+pub fn convert_range_to_inclusive<R>(range: R) -> Result<RangeInclusive<u32>, ConvertRangeError>
```
Kerollmops, line 25: Indeed the code is not pretty but it is very useful, thank you 😊

Kerollmops (approved):
Looks perfect to me, thank you very much for the hard work and patience 🐢

PR #299, opened 2024-11-12
Bump the MSRV to v1.67

Kerollmops:
To make the #297 CI pass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.66.0
+          - 1.67.0
     env:
       RUSTFLAGS: "-C target-cpu=native -C opt-level=3"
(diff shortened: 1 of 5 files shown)


Dr-Emann:
Could also consider [committing the lock file](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html)

Kerollmops:
> Could also consider [committing the lock file](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html)

I prefer to have errors so that I know when I need to upgrade the dependencies. I mean, there is maybe a better way to know that 🤔

Dr-Emann:
Can have a CI task that tries `cargo update` first, but isn't fatal if it fails, something like: https://github.com/RoaringBitmap/croaring-rs/blob/96fa65071c2fdbc6ab285613f7c14802cb104c7d/.github/workflows/rust.yml#L64-L66

PR #305, opened 2024-12-08
fix(no_std): use `core::slice` in `bitmap_store.rs`

GZTimeWalker:
fix https://github.com/RoaringBitmap/roaring-rs/issues/304

one more thing, why the ci did not find this error?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,88 +34,59 @@ jobs:
           path: "benchmarks/real-roaring-datasets"
 
       - name: Initialize rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
+        if: matrix.rust != 'nightly'
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
-          override: true
           components: rustfmt, clippy
 
-      - name: Install miri
+      - name: Install nightly toolchain
         uses: actions-rs/toolchain@v1
         if: matrix.rust == 'nightly'
         with:
-          profile: minimal
-          toolchain: nightly
-          override: true
-          components: miri
+          toolchain: ${{ matrix.rust }}
+          components: rustfmt, clippy, miri
 
-      - name: miri setup
-        uses: actions-rs/cargo@v1
+      - name: Setup miri
         if: matrix.rust == 'nightly'
-        with:
-          command: miri
-          args: setup
+        run: cargo miri setup
 
       - name: Fetch
-        uses: actions-rs/cargo@v1
-        with:
-          command: fetch
+        run: cargo fetch
 
       - name: Build
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --all-targets
+        run: cargo build --all-targets
 
       - name: Check
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --all-targets -- -D warnings
+        run: cargo clippy --all-targets -- -D warnings
 
       - name: Check formatting
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: -- --check
+        run: cargo fmt --all -- --check
 
       - name: Test
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --features serde
+        working-directory: roaring
+        run: cargo test --features serde
 
       - name: Test bit endian
-        uses: actions-rs/cargo@v1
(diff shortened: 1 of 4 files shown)


Kerollmops:
Hey @GZTimeWalker 👋

Could you please add a small test and CI check to ensure we don't break it again?

Thank you and have a nice week 🏝️

GZTimeWalker:
Hey, let me try.

GZTimeWalker:
btw, you can upgrade the deps: `criterion` to `0.5` and `zip` to `2.2`

GZTimeWalker:
Here is things we found:

- [actions-rs](https://github.com/actions-rs) is archived since 2023
- According to [serde](https://github.com/serde-rs/serde/blob/46e9ecfcdd5216929ebcf29e76adc072412c5380/.github/workflows/ci.yml#L54), the core issue is running `cargo test` in the root dir, which will still pass the default feature to inner packages.
- We are working on to accelerate to ci process, but btw, is the ~3h benchmark necessary? I suggest to add a [github action inputs](https://docs.github.com/zh/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#inputs) to run it when releasing or run it manually.

Kerollmops:
> * [actions-rs](https://github.com/actions-rs) is archived since 2023

Yup, but I didn't find the time to check for an alternative. It is working great for now, so I prefer to keep it like that rather than invest time in it right now.

> * According to [serde](https://github.com/serde-rs/serde/blob/46e9ecfcdd5216929ebcf29e76adc072412c5380/.github/workflows/ci.yml#L54), the core issue is running `cargo test` in the root dir, which will still pass the default feature to inner packages.

Could you please run a `cargo test -p roaring` in the CI? Do you think it can fix this issue?

> * We are working on to accelerate to ci process, but btw, is the ~3h benchmark necessary? I suggest to add a [github action inputs](https://docs.github.com/zh/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#inputs) to run it when releasing or run it manually.

Can we propose a CI file for this, please? Modify the README to explain how to run the benchmarks CI (it will be available for the repo's owners only, obviously).

GZTimeWalker:
For optional benchmarks, I suggest you to check how to launch it by pr tags, I think it can be implemented by add a `bench` tag for a pr which need to do a benchmark.

But I don't know how to do it, maybe post a help wanted or search for other's action file.

YanWQ-monad:
> > * We are working on to accelerate to ci process, but btw, is the ~3h benchmark necessary? I suggest to add a [github action inputs](https://docs.github.com/zh/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#inputs) to run it when releasing or run it manually.
> 
> Can we propose a CI file for this, please? Modify the README to explain how to run the benchmarks CI (it will be available for the repo's owners only, obviously).

I think before talking about improving the benchmark CI, should we make it clear what the benchmark is used for? For now, I can't see anything except the benchmark does run in CI. I mean, is it actually be used when deciding to merge a PR, or will it send a notification if a commit causes regression? And even if it is, the result of benchmark are not used to compare with the previous or subsequent commits, which may make the result hard to interpret.

In my opinion, for example, if we want to monitor the improvement or regression, it would be OK to run benchmark every time (the CI runs at no cost, though it takes time). But we can't run the benchmark only once just like now, because it can't tell if an improvement or regression happens. So we need the baseline result for comparison, it may come from running the benchmark on base commit (but it takes 2x time), or from pre-saved results (e.g. save to Actions Cache?), which may need some changes to the CI process.

Anyways, I think the improvement of benchmark CI is a little bit off-topic to this PR, would it be great to open an issue for it?

GZTimeWalker:
I agree to discuss the use of benchmark in another issue, and this PR has done its job.

It now correctly checks for compilation issues in all feature cases and runs all test tasks.

GZTimeWalker:
`cargo test -p roaring` is a good idea for test the package.

image

GZTimeWalker:
Any further on this? The most important thing now is that I need this library to compile properly under no_std.

Kerollmops (approved):
Hey @GZTimeWalker 👋

Thank you very much for the changes. It's very cool to fix this kind of issues.

> I agree to discuss the use of benchmark in another issue, and this PR has done its job.

Would it be possible to open an issue about this and the ideas of @YanWQ-monad, please?

In the meantime, I plan to bump the roaring crate version and release it today.

Have a nice day 🐢

PR #311, opened 2025-01-31
chore(deps): replace `bincode` with `postcard`

GZTimeWalker:
motivations:

- `bincode` has not released a new version for over a year, and many plans for v2.0 have yet to be implemented.
- `postcard` is in better maintenance condition and is sufficiently suitable for testing purposes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,11 +1,10 @@
 [workspace]
-members = ["roaring", "benchmarks"]
+members = [ "benchmarks", "roaring" ]
 resolver = "2"
 
 [workspace.dependencies]
 roaring = { path = "roaring" }
 
-bincode = "1.3.3"
 bytemuck = "1.21.0"
 byteorder = "1.5.0"
 criterion = "0.5"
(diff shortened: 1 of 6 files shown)


Kerollmops:
Hey @GZTimeWalker 👋

Thank you for your contribution. However, I am not sure why we depend on bincode. Replacing the bincode dependency by postcard is breaking and I would prefer to entirely remove it than replacing it. Also, replacing Bincode, which has 3.6M downloads by month, with Postcard, which has 658k downloads by month, is probably not the best idea.

GZTimeWalker:
Well, replacing the bincode dependency by postcard is NOT a breaking change, which is only used to test serde, and users will not install bincode or postcard when they use roaring.

GZTimeWalker:
You can take a look at https://github.com/RoaringBitmap/roaring-rs/blob/main/roaring/Cargo.toml , it is a dev-dependence there.

And also there is some fix for clippy CI.

Kerollmops:
Thank you for the changes @GZTimeWalker and sorry for the quick PR closing again!

GZTimeWalker:
> Thank you for the changes @GZTimeWalker and sorry for the quick PR closing again!

Never mind, thanks!

GZTimeWalker:
Well, bincode released v2.0 today in https://github.com/bincode-org/bincode

It seems that they continue to maintain the repo 4 days before.

Issue #312, opened 2025-03-19
`advance_to` not implemented for `roaring::treemap::Iter`


sdd:
I have a use case where I need the `advance_to` method, which is present on `RoaringBitmap`'s `Iter`, but I want to use it from a `RoaringTreemap`. This doesn't seem to exist - I'm happy to contribute an implementation, unless there a reason why this is not implemented already that I'm unaware of. Would you be open to this?

Additionally, I looked into implementing a struct outside of this library that would provide this functionality, but it's not possible to implement this easily because `roaring::treemap::iter::BitmapIter` is not publicly exported. Is this just an oversight? It's not currently possible to store the result of the public [RoaringTreemap::bitmaps](https://docs.rs/roaring/latest/roaring/treemap/struct.RoaringTreemap.html#method.bitmaps) method in a struct because of this.

sdd:
> Ho, yes, please contribute. The only reason it doesn't exist is that we didn't spend time on it.

With pleasure, I'm only too happy to give something back to this great library! :-)

> I think it is an oversight indeed. All publicly accessible struct/types should be available. Do you know about a certain edition, compiler parameter or CI setup we can use to make sure the CI fails in case we miss that again?

I don't, off the top of my head - but I'll see what I can find and incorporate it into a PR with the fix, if I find something.

sdd:
Looks like there are no Clippy lints that do this. But [dylint](https://github.com/trailofbits/dylint) could do it. I've not written a dylint linter before - I'll dig a bit deeper to see how hard it would be for this case.

sdd:
It works only if I add a reference to the returned type to the doc comment, and also only if I run `cargo rustdoc -p roaring`. It seems that the `cargo clippy -p roaring --all-targets --no-default-features -- -D warnings` command that runs in the GitHub action does not check rustdoc lints:

```
❯ cargo rustdoc -p roaring
 Documenting roaring v0.10.10 (/Users/scott/projects/roaring-rs/roaring)
warning: public documentation for `bitmaps` links to private item `BitmapIter`
   --> roaring/src/treemap/iter.rs:247:21
    |
247 |     /// Returns a [`BitmapIter`].
    |                     ^^^^^^^^^^ this item is private
    |
    = note: this link will resolve properly if you pass `--document-private-items`
    = note: `#[warn(rustdoc::private_intra_doc_links)]` on by default

warning: `roaring` (lib doc) generated 1 warning
[...]

PR #314, opened 2025-03-21
Add `advance_to` and `advance_back_to` to `Treemap`'s  `Iter`

sdd:
Partially addresses https://github.com/RoaringBitmap/roaring-rs/issues/312.

I can see now why this was not already implemented 😅 

I'm aware that I'm a newcomer to contributing to this repo so please excuse me if I've gone way off-piste. Implementing this efficiently by permitting the iterator to skip entire bitmaps for values of `n` that are beyond the range of the current bitmap led to a substantial refactoring so that `Iter` has access to the values that it needs to do this.

I've left out `IntoIter` for now, and left `fold` and `rfold` implementations as todos, because I wanted to get feedback on the approach first.

I've duplicated the tests from bitmap's `advance_to` and added a couple more to exercise the bitmap-skipping code but the suite is certainly not yet exhaustive, and again I'll extend the coverage here if you think I'm not completely on the wrong track.

### DIFF
--- a/roaring/src/treemap/iter.rs
+++ b/roaring/src/treemap/iter.rs
@@ -1,5 +1,6 @@
 use alloc::collections::{btree_map, BTreeMap};
 use core::iter;
+use std::iter::Peekable;
 
 use super::util;
 use crate::bitmap::IntoIter as IntoIter32;
@@ -11,10 +12,24 @@ struct To64Iter<'a> {
     inner: Iter32<'a>,
 }
 
+impl To64Iter<'_> {
+    fn advance_to(&mut self, n: u32) {
+        self.inner.advance_to(n)
+    }
+
+    fn advance_back_to(&mut self, n: u32) {
+        self.inner.advance_back_to(n)
+    }
+}
+
 impl Iterator for To64Iter<'_> {
     type Item = u64;
     fn next(&mut self) -> Option<u64> {
         self.inner.next().map(|n| util::join(self.hi, n))
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.inner.size_hint()
     }
 
     #[inline]
@@ -42,8 +57,8 @@ impl DoubleEndedIterator for To64Iter<'_> {
     }
 }
 
-fn to64iter<'a>(t: (&'a u32, &'a RoaringBitmap)) -> To64Iter<'a> {
-    To64Iter { hi: *t.0, inner: t.1.iter() }
+fn to64iter(t: (u32, &RoaringBitmap)) -> To64Iter<'_> {
+    To64Iter { hi: t.0, inner: t.1.iter() }
 }
 
 struct To64IntoIter {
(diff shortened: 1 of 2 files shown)


Dr-Emann:
Hmm, It's unfortunate that we have to skip entries in the BTreeMap one by one, ideally we should be able to skip right to the right start entry (O(log(N))). I think ideally, [btree_map::Range](https://doc.rust-lang.org/std/collections/btree_map/struct.Range.html) would allow you to re-narrow the range. I think maybe storing the reference to the BTreeMap, and a next begin/end, you could use `BTreeMap::range(..).next()` to skip ahead to the right place when needed.

sdd:
Thanks for the pointer, I'll refactor with that in mind 👍🏼

sdd:
Skipping directly to the right start entry seems pretty straightforward to do with `Range`. The problem with doing that though is updating the size hint. We need to iterate over the intervening entries in order to accumulate all of the counts to correctly reduce the size hint.  I don't see a way around this whilst keeping the size hint correct.

Dr-Emann:
Deja vu all over again: 😅

https://github.com/RoaringBitmap/roaring-rs/issues/294

I think I'd argue to rip out the precomputed size hint here too.

sdd:
I'll update this PR to dynamically calculate the size_hint so we can see what kind of impact that has.

sdd:
Hey @Kerollmops / @Dr-Emann - I've updated to dynamically calc the size hint so that we can efficiently advance the treemap. LMK if this is a reasonable approach. I'm not sure how to interpret the criterion results in the benchmarks - is there an easy way to compare to a baseline without manually comparing to a recent run on master?

Kerollmops:
Hey @sdd 👋

Thank you for the clean work.

> I've updated to dynamically calc the size hint so that we can efficiently advance the treemap. LMK if this is a reasonable approach.

I think that's good, I hope it's not too long to process. The `Iterator::size_hint` method [is usually used for optimizations](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.size_hint). According to the documentation (copied/pasted below), it is forbidden to provide an estimated size; the size hint must be exact.

The current implementation seems to be fast enough. It only sums the container sizes. May be enough.

> `size_hint()` is primarily intended to be used for optimizations such as reserving space for the elements of the iterator, but must not be trusted to e.g., omit bounds checks in unsafe code. An incorrect implementation of `size_hint()` should not lead to memory safety violations.
>
> That said, the implementation should provide a correct estimation, because otherwise it would be a violation of the trait’s protocol.
>
> The default implementation returns <code>(0, [None](https://doc.rust-lang.org/stable/std/option/enum.Option.html#variant.None)</code>) which is correct for any iterator.

Let's merge this and release it like that 💪
Have a nice day ☕

Kerollmops (approved):

sdd:
Thanks @Kerollmops - I wasn't expecting a merge yet, I was more looking for feedback on the approach. The tree map iterator fold and rfold methods are still in there as `todo!`s 😅

I better get cracking on implementations for them then!

Kerollmops:
Ho! Noooo! My bad! 😞 I need to setup Clippy to warn about todos.

Kerollmops:
@sdd, I just yanked the 0.10.11 version, deleted the tag, deleted the release, and apologize again to you 😞
I'll set a new CI with Clippy warning about `todo!`s so I don't repeat the mistake.
Would it be possible to keep the PR as a draft while it's not ready to be merged, please?

sdd:
Of course, sorry for the confusion. I'll get those todo's sorted, as well as the IntoIter implementation, over the coming few days. 👍

PR #319, opened 2025-04-04
Remove `todo!` Implementations for Treemap `Iter`'s `fold` and `rfold`

sdd:
I think it makes sense to just rely on the default implementations for `fold` and `rfold` for `Iter`. Providing custom implementations for them would likely just be mostly duplicates of the code in `next` and `next_back` anyway with no optimisations over the defaults.


Kerollmops (approved):
That's a good idea and that's simpler to review and maintain anyway 😅

PR #321, opened 2025-04-17
Add fuzzing to compare against croaring-rs

Dr-Emann:
This also adds some extra validation which will reject a few more incorrect bitmaps when deserializing, to reject inputs also rejected by croaring, and some tests mirroring the unit tests for corrupted deserialization added in https://github.com/RoaringBitmap/CRoaring/pull/665

I figure this might also be helpful for use in #320 (@lucascool12 FYI).

This has been running for a while on my machine (about 300 million executions over 19 hours) and hasn't found anything new in that time.


Kerollmops:
Wow! That is impressive work. Thank you very much!

Don't we want to run it in place of the current benchmarks in the CI for something like 30 minutes?
What do you think?

Dr-Emann:
@Kerollmops gave a try at adding 30 minutes of fuzzing to CI instead of the benchmarks, not a github actions expert.

Kerollmops:
Thank you very much @Dr-Emann 🙏

It seems to me that it worked perfectly in the CI: _Continuous integration / fuzz pull_request Successful in 31m_ ✅ 

Are the following lines expected?
```
WARNING: Failed to find function "__sanitizer_acquire_crash_state".
WARNING: Failed to find function "__sanitizer_print_stack_trace".
WARNING: Failed to find function "__sanitizer_set_death_callback".

```

roaring/src/bitmap/serialization.rs
Kerollmops, line 1: Thank you for those improvements, too!

Kerollmops (approved):

Dr-Emann:
Yeah, that's expected. It's fuzzing without a sanitizer (-s none), because it's faster, and we don't expect any memory unsafety issues, the fuzzer has hooks into sanitizers, but it can't use them if we're not running with a sanitizer

PR #325, opened 2025-06-17
Change the API of the push method to return a result

Kerollmops:
This PR improves the API to ensure we don't make the same mistake of forgetting to check the boolean returned by the `push` method. We made multiple mistakes previously at Meilisearch, so I think the best approach is to deprecate it and introduce a new `try_push` method that returns a `Result`.

### DIFF
--- a/roaring/src/lib.rs
+++ b/roaring/src/lib.rs
@@ -33,6 +33,16 @@ pub mod treemap;
 pub use bitmap::RoaringBitmap;
 pub use treemap::RoaringTreemap;
 
+/// An error type that is returned when a push in a bitmap did not succeed.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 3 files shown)


roaring/src/lib.rs
```diff
@@ -33,6 +33,16 @@ pub mod treemap;
 pub use bitmap::RoaringBitmap;
 pub use treemap::RoaringTreemap;
 
+/// An error type that is returned when a push in a bitmap did not succeed.
```
irevoire, line 36:
```suggestion
/// An error type that is returned when a try_push in a bitmap did not succeed.
```

irevoire (changes requested):

irevoire (approved):
Perfecto

Dr-Emann:
Probably too late, but I think you could just mark the push function with `#[must_use]`

PR #331, opened 2025-07-10
use a dynamic bytes swizzle for `swizzle_to_front`

### LINKED ISSUE #207: Remove static jump table from `swizzle_to_front`
[This monstrosity](https://github.com/RoaringBitmap/roaring-rs/blob/95a092528e17c565dede6ba31659d8a70d8e90a5/src/bitmap/store/array_store/vector.rs#L507) needs to go when possible.

Blocked by rust-lang/portable-simd#11

Dr-Emann:
Fixes #207

As the comments say, unfortunately, this leads to pretty suboptimal code unless the std library is compiled with the required target feature, and the default x86_64 target (except for macos) is not compiled with the required target feature.

See [godbolt](https://rust.godbolt.org/z/q99ajvYv6) showing assembly for: default x86_64 (bad assembly), aarch64 (good assembly), and x86_64 macos (good assembly, which should also be the assembly for `-Zbuild-std`)

We _could_ use a little unsafe and do something like:
```rust
#[cfg(target_feature = "ssse3")]
{
    use core::arch::x86_64::{__m128i, _mm_shuffle_epi8};
    let val_m128 = __m128i::from(val_convert);
    let swizzle_m128 = __m128i::from(swizzle_idxs);
    let swizzled_m128 = unsafe { _mm_shuffle_epi8(val_m128, swizzle_m128) };
    return u16x8::from(swizzled_m128);
}
```

to allow [optimal code generation](https://rust.godbolt.org/z/MPqxGj7n3) with just a new enough target in the caller without requiring recompiling std.

### DIFF
--- a/roaring/src/bitmap/store/array_store/vector.rs
+++ b/roaring/src/bitmap/store/array_store/vector.rs
@@ -498,9 +498,6 @@ fn simd_merge_u16(a: Simd<u16, 8>, b: Simd<u16, 8>) -> [Simd<u16, 8>; 2] {
 // > of the standard library, so `cargo build -Zbuild-std` may be necessary
 // > to unlock better performance, especially for larger vectors.
 // > A planned compiler improvement will enable using `#[target_feature]` instead.
-//
-// Specifically, e.g. the default `x86_64` target does not enable ssse3, so this may be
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
Amazing work! Thank you for this 👍

PR #335, opened 2025-07-23
Fuzz iterators and fix bugs

Dr-Emann:
Seems to at least be able to pass for a while on fuzzing after these changes.


Dr-Emann:
Will fail until #336

PR #336, opened 2025-07-23
fix: clippy findings from updated msrv

Dr-Emann:
Bumping the minimum rust version in #333 let clippy know we can use functions introduced before our msrv, so there's some findings it has.


Kerollmops (approved):
Nice catch 🐟

PR #346, opened 2026-01-18
feat(iter): Add next_many() for batch iteration

Zeutschler:
## Summary

Add `next_many(&mut self, dst: &mut [u32]) -> &[u32]` method to `Iter` and `IntoIter` for efficient batch extraction of bitmap values into a user-provided buffer.

## Motivation

When iterating over large bitmaps, calling `next()` repeatedly incurs significant per-element overhead. The `next_many()` method extracts multiple values at once, enabling:

- Reduced function call overhead
- Better cache locality with contiguous buffer writes
- ILP-friendly processing of batched results

This API mirrors the `next_many()` method available in [CRoaring](https://github.com/RoaringBitmap/CRoaring) and the [Go implementation](https://github.com/RoaringBitmap/roaring) of RoaringBitmap.

## Performance

| Benchmark | `next()` | `next_many()` | Speedup |
|-----------|----------|---------------|---------|
| Dense (1M values, bitmap storage) | 19.02ms | 6.16ms | **3.09x** |
| Sparse (10K values, array storage) | 1.67ms | 159.58µs | **10.46x** |

## API

```rust
let mut iter = bitmap.iter();
let mut buf = [0u32; 1024];
loop {
    let out = iter.next_many(&mut buf);
    if out.is_empty() { break; }
    // Process out
}
```

Returns an immutable slice of the `dst` buffer. Returns an empty slice when the iterator is exhausted.

## Implementation Details

- **Store::Iter**: Direct slice copy for Array/Vec storage, bit extraction for Bitmap storage, run expansion for Interval stores
- **Container::Iter**: Uses u16 buffer internally, combines with container key to produce u32 values
- **Bitmap::Iter/IntoIter**: Handles front/back iterators from `DoubleEndedIterator` and container transitions

## Checklist

- [x] Tests pass (`cargo test`)
- [x] No breaking changes
- [x] Backward compatible

### DIFF
--- a/roaring/src/bitmap/container.rs
+++ b/roaring/src/bitmap/container.rs
@@ -473,6 +473,34 @@ impl Iter<'_> {
             .next_range_back()
             .map(|r| util::join(self.key, *r.start())..=util::join(self.key, *r.end()))
     }
+
+    /// Read multiple values from the iterator into `dst`.
+    /// Returns the number of values read.
+    ///
+    /// This can be significantly faster than calling `next()` repeatedly.
+    pub(crate) fn next_many(&mut self, dst: &mut [u32]) -> usize {
+        // Use a temporary u16 buffer for the inner iterator
+        const BUF_SIZE: usize = 256;
+        let mut buf = [0u16; BUF_SIZE];
+
+        let key = self.key;
+        let mut count = 0;
+
+        while count < dst.len() {
+            let remaining = dst.len() - count;
+            let to_read = remaining.min(BUF_SIZE);
+            let n = self.inner.next_many(&mut buf[..to_read]);
+            if n == 0 {
+                break;
+            }
+            for i in 0..n {
+                dst[count + i] = util::join(key, buf[i]);
+            }
+            count += n;
+        }
+
+        count
+    }
 }
 
 impl fmt::Debug for Container {
--- a/roaring/src/bitmap/container.rs
+++ b/roaring/src/bitmap/container.rs
@@ -478,7 +478,7 @@ impl Iter<'_> {
     /// Returns the number of values read.
     ///
     /// This can be significantly faster than calling `next()` repeatedly.
-    pub(crate) fn next_many(&mut self, dst: &mut [u32]) -> usize {
+    pub(crate) fn next_many<'a>(&mut self, dst: &'a mut [u32]) -> &'a [u32] {
         // Use a temporary u16 buffer for the inner iterator
         const BUF_SIZE: usize = 256;
         let mut buf = [0u16; BUF_SIZE];
@@ -489,17 +489,17 @@ impl Iter<'_> {
         while count < dst.len() {
             let remaining = dst.len() - count;
             let to_read = remaining.min(BUF_SIZE);
-            let n = self.inner.next_many(&mut buf[..to_read]);
-            if n == 0 {
+            let out_len = self.inner.next_many(&mut buf[..to_read]).len();
+            if out_len == 0 {
                 break;
             }
-            for i in 0..n {
+            for i in 0..out_len {
                 dst[count + i] = util::join(key, buf[i]);
             }
-            count += n;
+            count += out_len;
         }
 
-        count
+        &dst[..count]
     }
 }
 
--- a/roaring/src/bitmap/container.rs
+++ b/roaring/src/bitmap/container.rs
@@ -475,7 +475,7 @@ impl Iter<'_> {
     }
 
     /// Read multiple values from the iterator into `dst`.
-    /// Returns the number of values read.
+    /// Returns a mutable slice of `dst` that contains the read values.
     ///
     /// This can be significantly faster than calling `next()` repeatedly.
     pub(crate) fn next_many<'a>(&mut self, dst: &'a mut [u32]) -> &'a mut [u32] {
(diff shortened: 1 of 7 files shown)


roaring/src/bitmap/iter.rs
```diff
@@ -331,6 +331,82 @@ impl Iter<'_> {
...
+    /// assert_eq!(n, 32);
+    /// assert_eq!(buf[0], 0);
+    /// assert_eq!(buf[31], 31);
+    ///
+    /// // Iterate remainder
+    /// let n = iter.next_many(&mut buf);
+    /// assert_eq!(n, 32);
+    /// assert_eq!(buf[0], 32);
+    /// ```
+    pub fn next_many(&mut self, dst: &mut [u32]) -> usize {
```
Kerollmops:
```suggestion
    pub fn next_many<'a>(&mut self, dst: &'a mut [u32]) ->&'a [u32] {
```

roaring/src/bitmap/iter.rs
```diff
@@ -419,6 +495,82 @@ impl IntoIter {
...
```
Kerollmops:
```suggestion
    pub fn next_many<'a>(&mut self, dst: &'a mut [u32]) -> &'a [u32] {
```

Kerollmops (commented):
Hey @Zeutschler 👋

That's a great addition to the crate 🎉 

I took a look at the usage, and it seems to me that returning an `&[u32]` would be preferable, as I often see you use it like `&[..n]` and it is a source of errors to me. If anyone wants the number, they can use the `[]::len()` method, or the `[]::is_empty()` method to check whether it's empty.

Would you mind changing the signature of the `Iter::next_many` method and also the internal methods, too? Would it also be possible to add more proptests to check if the interaction between the classic `Iterator::next` and the new `next_many` methods, please?

I am wondering: if we don't want to introduce a trait rather than raw methods, we could implement them on different structs, and people can use trait constraints instead 🤔 What do you think?

Thank you and have a nice day 🌵

roaring/src/bitmap/iter.rs
Zeutschler: Hi @Kerollmops, yes that makes more sense, more flexible. Go for it. Thx for your great work.👍

Kerollmops: Hey @Zeutschler, do you plan to work on this or do you want me to do it?

Zeutschler: I'm very busy for the next months. So, if you can do it that would be great. Thx... (Sorry)

Kerollmops: Ok, no problem, I'll update your PR then. I hope I can 👀

Dr-Emann:
For CRoaring-rs, I did [next_many(&mut self, dst: &mut [u32]) -> usize](https://docs.rs/croaring/latest/croaring/bitmap/struct.BitmapIterator.html#method.next_many).

I used `-> usize` for consistency with the `Read` trait, but thinking about the differences, here's what I came up with:

- Before NLL, returning a slice could be more painful, but that's no longer a concern
- In a trait, especially one often used so close to unsafe, it's nice that the API requires the bytes are written to the provided slice, without an offset, etc. This doesn't really apply in this case.

I think this should return `-> &'a mut [u32]`, rather than an immutable slice, though, rather than unnecessarily losing mutability.

roaring/src/bitmap/store/bitmap_store.rs
```diff
@@ -691,6 +691,55 @@ impl<B: Borrow<[u64; BITMAP_LENGTH]>> BitmapIter<B> {
         let index = 63 - index_from_left;
         Some(64 * key_back + index)
     }
+
+    /// Read multiple values from the iterator into `dst`.
+    /// Returns the number of values read.
```
Dr-Emann: This comment isn't accurate anymore

roaring/src/bitmap/container.rs
```diff
@@ -473,6 +473,34 @@ impl Iter<'_> {
...
     }
+
+    /// Read multiple values from the iterator into `dst`.
+    /// Returns the number of values read.
+    ///
+    /// This can be significantly faster than calling `next()` repeatedly.
+    pub(crate) fn next_many<'a>(&mut self, dst: &'a mut [u32]) -> &'a [u32] {
+        // Use a temporary u16 buffer for the inner iterator
+        const BUF_SIZE: usize = 256;
+        let mut buf = [0u16; BUF_SIZE];
```
Dr-Emann: In the C version, the container "next_many" implmentation instead takes (basically) a `dst: &mut [u32]` and a `high_16_bits: u32`, to avoid having to do all the copying from a separate buffer of u16s:
https://github.com/RoaringBitmap/CRoaring/blob/4c219986438ed06c7ee26b70ccb2ed4982887e5a/src/containers/containers.c#L420-L424

Kerollmops:
Hey @Dr-Emann 👋

Long time no see! Thanks again for the review. I changed the API based on your feedback and modified the implementation of the `next_many` container methods to avoid copying too much between the u16 arrays and the u32 array output. I used the `util::join` method. What do you think?

After these improvements, @Zeutschler, you would probably want to rerun your quick benchmarks 👀

Have a nice day 🌵

PR #347, opened 2026-02-16
Remove SIMD LaneCount, SupportedLaneCount with upstream changes

xangelix:
Not sure exactly when this will be part of a versioned release? As such this is marked draft. This contains the required changes for SIMD compilation on the latest nightly toolchain.

See https://github.com/rust-lang/rust/commit/b71ff51277dbd5907a7a8089b5d926a02e53d44d.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
           - nightly
           # When changing this value don't forget to change the `package.rust-version` field in
           # `roaring/Cargo.toml`!!!
-          - 1.82.0
+          - 1.90.0
     env:
       RUSTFLAGS: "-C target-cpu=native -C opt-level=3"
 
@@ -64,7 +64,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.82.0
+          - 1.90.0
         features:
           - default
           - no-std
@@ -98,7 +98,7 @@ jobs:
         run: cargo test -p roaring --features serde
 
       - name: Test Benches
-        if: matrix.rust != '1.82.0' && matrix.features == 'default'
+        if: matrix.rust != '1.90.0' && matrix.features == 'default'
         run: cargo test -p benchmarks --benches
 
       - name: Test no default features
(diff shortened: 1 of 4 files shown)


Kerollmops (commented):
Nice 👌  Thank you very much @xangelix for the change. Let's see where it goes.

Can you tell me if I need to land your changes to crates.io so that people can use the `simd` feature with the latest rustc compile, please?

Kerollmops (changes requested):
Would you mind rebasing on #349, please? It's odd because it seems to me that your PR will fix nightly, but will break the stable CI...

xangelix:
I would be happy to rebase so your CICD can work! But to be clear, I don't think this PR breaks stable.

`portable_simd` is only a feature on nightly, so this comes down to which version of `nightly` you'd want to pin to based roughly on that linked commit above, https://github.com/rust-lang/rust/commit/b71ff51277dbd5907a7a8089b5d926a02e53d44d , you'd decide if you want to support **before or after** Jan 28th, 2026.

If you want to support nightly versions roughly *after*  `nightly-2026-01-28` then you would merge this, if you want to support roughly *before* `nightly-2026-01-28`, then you would wait on this.

---

Separately about the CICD,

If you check with `cargo tree`:

image

it's `tempfile` which is forcing the `getrandom` v0.4 dependency.

Precisely v3.25.0, https://github.com/Stebalien/tempfile/blob/master/CHANGELOG.md#3250 which was released 15 days ago https://crates.io/crates/tempfile/3.25.0.

Meaning that for 15 days now, your CICD would fail even if you ran your main branch on it. (So the CICD was already broken :frowning_face:)

Anyways, sorry for the ramble! I think this PR is just a matter of which nightly versions you might want to support. **So, it's not necessary to bump the MSRV for this PR if you don't want to.** Correct me if I'm wrong! :stuck_out_tongue:

Rebasing now...

Kerollmops (approved):
Looks good to me, the CI is passing now 👏

PR #348, opened 2026-02-16
chore: resolve all missing_const_for_fn clippy lints

xangelix:
24x missing_const_for_fn

### DIFF
--- a/roaring/src/bitmap/container.rs
+++ b/roaring/src/bitmap/container.rs
@@ -26,7 +26,7 @@ pub(crate) struct Iter<'a> {
 }
 
 impl Container {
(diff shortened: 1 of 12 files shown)


Kerollmops (approved):
Thank you very much, @xangelix, for the changes. As far as I can tell, only the `NonSortedIntegers::valid_until` method is a user-facing one (a public method). I think it's good to be merged 👍

xangelix:
Hi @Kerollmops ! You're welcome-- happy to contribute.

Both `RoaringBitmap::new()` and `RoaringTreemap::new()` are made const with this PR as far as I can tell? Which was the primary reason for this PR. Among other public methods.

PR #349, opened 2026-02-24
Bump MSRV to v1.90.0

Kerollmops:
Trying to fix the CI of #348 by bumping the version to v1.90.0.


Kerollmops:
Closed as it is a subset of #347.

PR #351, opened 2026-04-21
Add RoaringTreemap::contains_range and range_cardinality

dchristle:
`RoaringBitmap` has `contains_range` and `range_cardinality`; `RoaringTreemap` does not.

Empty-range conventions match `RoaringBitmap`: `contains_range` returns `true`, `range_cardinality` returns `0`.

New test file `treemap_range_checks.rs` mirrors `range_checks.rs`.

### DIFF
--- a/roaring/src/treemap/inherent.rs
+++ b/roaring/src/treemap/inherent.rs
@@ -352,13 +352,14 @@ impl RoaringTreemap {
         let (start_hi, start_lo) = util::split(start);
         let (end_hi, end_lo) = util::split(end);
 
-        let mut cardinality = 0u64;
-        for (&key, bitmap) in self.map.range(start_hi..=end_hi) {
-            let lo_start = if key == start_hi { start_lo } else { 0 };
-            let lo_end = if key == end_hi { end_lo } else { u32::MAX };
-            cardinality += bitmap.range_cardinality(lo_start..=lo_end);
-        }
-        cardinality
+        self.map
+            .range(start_hi..=end_hi)
+            .map(|(&key, bitmap)| {
+                let lo_start = if key == start_hi { start_lo } else { 0 };
+                let lo_end = if key == end_hi { end_lo } else { u32::MAX };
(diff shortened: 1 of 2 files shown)


roaring/src/treemap/inherent.rs
```diff
@@ -272,6 +272,95 @@ impl RoaringTreemap {
...
+        let (start_hi, start_lo) = util::split(start);
+        let (end_hi, end_lo) = util::split(end);
+
+        let mut cardinality = 0u64;
+        for (&key, bitmap) in self.map.range(start_hi..=end_hi) {
+            let lo_start = if key == start_hi { start_lo } else { 0 };
+            let lo_end = if key == end_hi { end_lo } else { u32::MAX };
+            cardinality += bitmap.range_cardinality(lo_start..=lo_end);
+        }
+        cardinality
```
Kerollmops: Could we convert this to a map + sum to benefit from auto-vectorization optimizations, maybe?

Kerollmops (changes requested):
Thank you very much for this PR. I see that Clippy may have been auto-updated and is triggering an error. Can you fix it quickly, please?

Overall, the PR looks good. I just have a small improvement proposal. Can you address it?

roaring/src/treemap/inherent.rs
Kerollmops: I did it!

Kerollmops (approved):
The CI issue doesn't seems related to your code so I'll merge it. Thanks for the help!

dchristle:
Thanks for the review and merge. Yes, the Clippy warning was on existing code this PR didn’t touch, so I think a newer Clippy version likely started flagging it.

FWIW I checked the `for` loop vs `map(...).sum()` with a local `--release` disassembly, and for this case they compiled equivalently on the tested target, so the difference looks stylistic rather than performance-driven. `BTreeMap::range(...)` already makes the loop iterator-driven, and the reduction is already obvious to the compiler.

PR #355, opened 2026-05-07
Implement quad search

### LINKED ISSUE #234: Optimization: investigate faster search in array
Some ideas for improving the binary search for array stores:
https://dirtyhandscoding.wordpress.com/2017/08/25/performance-comparison-linear-search-vs-binary-search/

IMO the two should be combined - binary search should be used to narrow down the range to the point where linear search is faster, then continue with that.

Kerollmops:
This PR implements the SIMD-based Quad search described by [Daniel Lemire in this blog post](https://lemire.me/blog/2026/04/27/you-can-beat-the-binary-search/), and fixes #234.

### DIFF
--- a/roaring/src/bitmap/store/array_store/vector.rs
+++ b/roaring/src/bitmap/store/array_store/vector.rs
@@ -548,3 +548,118 @@ pub fn swizzle_to_front(val: u16x8, bitmask: u8) -> u16x8 {
     let swizzled: u8x16 = val_convert.swizzle_dyn(swizzle_idxs);
     u16x8::from_ne_bytes(swizzled)
 }
+
+#[inline]
+pub fn quad_contains(slice: &[u16], val: u16) -> bool {
+    const GAP: usize = u16x8::LEN * 2;
+
+    let (chunks, remaining) = slice.as_chunks::<GAP>();
+
+    if chunks.is_empty() {
+        return match remaining.iter().copied().find(|v| *v >= val) {
+            Some(v) => v == val,
+            None => false,
+        };
+    }
+
+    let num_blocks = chunks.len();
+    let mut base = 0;
+    let mut n = num_blocks;
+    while n > 3 {
+        let quarter = n >> 2; // equivalent to n / 4
+
+        let k1 = chunks[base + quarter][GAP - 1];
+        let k2 = chunks[base + 2 * quarter][GAP - 1];
+        let k3 = chunks[base + 3 * quarter][GAP - 1];
+
+        let c1 = (k1 < val) as usize;
+        let c2 = (k2 < val) as usize;
+        let c3 = (k3 < val) as usize;
+
+        base += (c1 + c2 + c3) * quarter;
+        n -= 3 * quarter;
+    }
+
+    while n > 1 {
+        let half = n >> 1; // equivalent to n / 2
+        base = if chunks[base + half][GAP - 1] < val { base + half } else { base };
+        n -= half;
+    }
+
+    let lo = if chunks[base][GAP - 1] < val { base + 1 } else { base };
+
+    if lo < num_blocks {
+        let ndl = u16x8::splat(val);
+        // I would love to work with arrays here...
+        let v0 = u16x8::from_slice(&chunks[lo][..GAP / 2]);
+        let v1 = u16x8::from_slice(&chunks[lo][GAP / 2..]);
+        return (v0.simd_eq(ndl) | v1.simd_eq(ndl)).any();
+    }
+
+    match slice.iter().copied().skip(num_blocks * GAP).find(|v| *v >= val) {
+        Some(v) => v == val,
+        None => false,
+    }
+}
+
+#[inline]
+pub fn quad_search(slice: &[u16], val: u16) -> Result<usize, usize> {
+    const GAP: usize = u16x8::LEN * 2;
+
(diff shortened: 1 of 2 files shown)


roaring/src/bitmap/store/array_store/vector.rs
```diff
@@ -548,3 +548,118 @@ pub fn swizzle_to_front(val: u16x8, bitmask: u8) -> u16x8 {
     let swizzled: u8x16 = val_convert.swizzle_dyn(swizzle_idxs);
     u16x8::from_ne_bytes(swizzled)
 }
+
+#[inline]
+pub fn quad_contains(slice: &[u16], val: u16) -> bool {
```
Kerollmops, line 553: I am wondering whether I should introduce this function, since it is used only once in the `ArrayStore::contains` method, or simply use the `quad_search(...).is_ok()` function that returns the position where the number is found or should be.

roaring/src/bitmap/store/array_store/vector.rs
```diff
@@ -548,3 +548,118 @@ pub fn swizzle_to_front(val: u16x8, bitmask: u8) -> u16x8 {
...
+        n -= half;
+    }
+
+    let lo = if chunks[base][GAP - 1] < val { base + 1 } else { base };
+
+    if lo < num_blocks {
+        let ndl = u16x8::splat(val);
+        // I would love to work with arrays here...
+        let v0 = u16x8::from_slice(&chunks[lo][..GAP / 2]);
+        let v1 = u16x8::from_slice(&chunks[lo][GAP / 2..]);
```
Kerollmops, line 649: This `from_slice` method panics if the slice is too small; we need to check the generated assembler to see if it can avoid this redundant check.

roaring/src/bitmap/store/array_store/vector.rs
```diff
@@ -548,3 +548,118 @@ pub fn swizzle_to_front(val: u16x8, bitmask: u8) -> u16x8 {
...
+        let base_index = lo * GAP;
+        return match (v0.simd_ge(ndl).first_set(), v1.simd_ge(ndl).first_set()) {
+            (Some(i), _) if v0[i] == val => Ok(base_index + i),
+            (Some(i), _) => Err(base_index + i),
+            (_, Some(i)) if v1[i] == val => Ok(base_index + GAP / 2 + i),
+            (_, Some(i)) => Err(base_index + GAP / 2 + i),
+            (None, None) => Err(slice.len()),
+        };
```
Kerollmops, line 657: I am wondering if this is the best approach to take to compute the position where the needle is/should be located, or if I can do something else, less expensive than two `simd_ge` plus two `first_set`?

Kerollmops:
Hey @Dr-Emann 👋

I know that you like this kind of SIMD-based code, so if you want and have time to review this, I would be glad you do 🙏  If you don't have time, no worries 👍

Have a nice day 🌵

roaring/src/bitmap/store/array_store/vector.rs
```diff
@@ -548,3 +548,118 @@ pub fn swizzle_to_front(val: u16x8, bitmask: u8) -> u16x8 {
...
+    }
+
+    match slice.iter().copied().skip(num_blocks * GAP).find(|v| *v >= val) {
+        Some(v) => v == val,
+        None => false,
+    }
+}
+
+#[inline]
+pub fn quad_search(slice: &[u16], val: u16) -> Result<usize, usize> {
```
Dr-Emann, line 606: At least on my M1 Pro Macbook, `quad_search` seems to always lose to just rust stdlib binary search. 

Image

Kerollmops: Yeah, have you tried benchmarking `quad_contains`? Because, I designed `quad_search` to be behave like the `binary_search` method and return the position where we found or must insert the item. I probably did it wrong or in an unoptimized way.

Thanks for the review anyway 😊

Dr-Emann: Quad Contains also seems to lose to binary search. If I remove the bounds checks, it's competitive at lower array sizes, but binary search wins again at larger array sizes.

Benchmark code at https://gist.github.com/Dr-Emann/558a3116f9cd2f984673ecaa73d76b61

Godbolt showing no panics in the unsafe bounds check removed implementation: https://rust.godbolt.org/z/369WxTqKo

image

All on an M1 Mac, interested if you get different results on x64.

Dr-Emann: There's a [comment](https://lemire.me/blog/2026/04/27/you-can-beat-the-binary-search/#comment-677450) on the blog that seems relevant:

> The remainder of the article uses the branchy std::binary_search as a baseline for comparison to branchless SIMD implementations, which is a poor representation of the performance difference between the scalar and SIMD algorithms.

The rust binary search implementation is really good.

Dr-Emann: @lemire any interest in looking into if we're missing anything important here?

lemire: @Dr-Emann I got my AI to do a Rust port,

https://github.com/lemire/rustquadsimd


It even did the experiments...

https://github.com/lemire/rustquadsimd#results


I told the AI to keep things really simple.

PR #360, opened 2026-07-29
fix: invalid treemap iter advance

silver-ymz:
The PR fixes following cases:

- After calling next(), jump forward 32 bits in the bitmap.
- After calling next_back(), jump backward in the bitmap.
- When the target bitmap does not exist, do not apply the invalid target's lower 32 bits to the next bitmap.
- Symmetric backward missing bitmap situation.

You can review the test cases for detailed information about these bugs.

### DIFF
--- a/roaring/src/bitmap/store/array_store/mod.rs
+++ b/roaring/src/bitmap/store/array_store/mod.rs
@@ -102,11 +102,10 @@ impl ArrayStore {
 
         let mut vec = Vec::with_capacity(bits_set as usize);
 
-        let chunks = bytes.chunks_exact(size_of::<Word>());
-        let remainder = chunks.remainder();
-        for (index, chunk) in chunks.enumerate() {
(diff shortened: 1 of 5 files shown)


silver-ymz:
@Kerollmops Could you help to review it? Thank you very much!

Kerollmops:
Hey @silver-ymz 👋

Thanks for the bug report and fix. Would it be possible to create two commits: one with the tests you added and another with the fix itself. Can you push them one after the other so that I can see the CI breaking on your first commit (the next tests failing) and the other become green as it fixes the tests, please?

Have a nice day 🌵

silver-ymz:
`cargo clippy` failed due to a rust version update and is unrelated to this PR. It seems we need to first make clippy happy.

Kerollmops:
Hey @silver-ymz 👋

You're right it's a clippy issue but can you fix it in a specific commit, please? It seems that it's just a matter of using `as_chunk` 👀

PR #363, opened 2026-08-12
Fix off-by-one that corrupts a bitmap in remove_smallest/remove_biggest (#359)

### LINKED ISSUE #359: `remove_smallest` / `remove_biggest` can corrupt a bitmap
```rust
use roaring::RoaringBitmap;

let mut b = RoaringBitmap::new();
b.insert_range(0..=2);
b.insert(4);
b.remove_smallest(3);
```

On a debug build this panics:

```
thread 'main' panicked at src/bitmap/store/interval_store.rs:966:19:
attempt to subtract with overflow
```

On a release build there is no error, but the bitmap is corrupted. After removing the 3 smallest values from `{0, 1, 2, 4}` the result should be `{4}`; instead `b.len()` returns `65537` and iterating the bitmap yields `[3, 4, 5, 6, 7, 8, 9, 10, 11, 12, ...]`.

`remove_biggest` has the same problem, e.g. building `{2, 4, 5, 6, 7, 8}` and calling `remove_biggest(5)`. It happens when the amount removed exactly consumes one of the bitmap's internal runs; `RoaringTreemap` is affected too.

Tested on roaring 0.11.4.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

youdie006:
Fixes #359.

### Problem

For a run/interval container, `IntervalStore::remove_smallest(n)` and `remove_biggest(n)` have an off-by-one at the interval where removal stops when `n` **exactly** equals that interval's run length:

- `remove_smallest` does `start += amount`, moving `start` to `end + 1`.
- `remove_biggest` does `end -= amount`, moving `end` to `start - 1`.

Either way the interval becomes inverted (`start > end`), and `run_len()` (`end - start + 1`) then underflows on the subtraction: a **debug-build panic** ("attempt to subtract with overflow"), or **silent corruption** in release. The reporter's example: `{0,1,2,4}.remove_smallest(3)` should yield `{4}` but returns a bitmap of length 65537.

### Fix

Drop the exactly-consumed interval instead of shrinking it:

- `remove_smallest`: `if last_interval.run_len() < amount` becomes `<=`, so an interval whose whole run is removed is dropped via `remove_to += 1` rather than shrunk.
- `remove_biggest`: `if last_interval.run_len() >= amount` becomes `>`, so the exactly-consumed interval falls through to the `drain(remove_to..)` that removes it, instead of shrinking its `end`.

### Test

Added `remove_smallest_exact_interval_boundary` and `remove_biggest_exact_interval_boundary` (the reporter's two cases at the `IntervalStore` level). Red-green verified with `cargo test -p roaring`: before the fix both assert against inverted intervals (`[3, 2]` / `[4, 3]`); after, they drop the whole interval. The full `roaring` test suite (including the property tests) passes; `rustfmt --check` and `clippy` are clean.

### DIFF
--- a/roaring/src/bitmap/store/interval_store.rs
+++ b/roaring/src/bitmap/store/interval_store.rs
@@ -261,7 +261,10 @@ impl IntervalStore {
             }
         }
         if let Some(last_interval) = last_interval {
-            if last_interval.run_len() < amount {
+            // `<=`, not `<`: when the amount left equals this interval's run
+            // length the whole interval is consumed and must be dropped, not
+            // shrunk to `start = end + 1` (an inverted interval) (#359).
+            if last_interval.run_len() <= amount {
                 remove_to += 1;
             } else {
                 last_interval.start += amount as u16;
@@ -285,7 +288,11 @@ impl IntervalStore {
             }
         }
         if let Some(last_interval) = last_interval {
-            if last_interval.run_len() >= amount {
+            // `>`, not `>=`: when the amount left equals this interval's run
+            // length the whole interval is consumed and is removed by the
+            // `drain` below; shrinking it would set `end = start - 1` (an
+            // inverted interval) (#359).
+            if last_interval.run_len() > amount {
                 remove_to += 1;
                 last_interval.end -= amount as u16;
             }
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
Hey @youdie006 👋

Thank you very much to you and @DRMacIver for the report and the bug fix. I'll release this fix as part of the next version release.

Have a nice day 🌵

PR #366, opened 2026-09-10
feat(retain): add `retain` method

HellButcher:
adds a `retain(&mut self, f: FnMut(u32) -> bool)` method to `RoaringBitmap` and `RoaringTreemap`.
It works similar to `Vec::retain`.


Kerollmops (changes requested):
Hey @HellButcher 👋

Thank you for the nice addition. Would you mind fixing the CI errors, please?

Kerollmops (changes requested):
Final little clippy issue now 👀 😊

PR #367, opened 2026-09-11
Optimize serialization for little-endian

Liblor:
Because the in-memory representation on little-endian targets matches the wire format, we can cast the slice to bytes and emit the container in a single `writer.write_all()` mirroring the CRoaring implementation. This speeds up the serialization 3-8x on a 11th Gen Intel i7-11850H.

Benchmark run: https://gist.github.com/Liblor/3e5cf40d3a79c831124d75ef69471850
CRoaring: https://github.com/RoaringBitmap/CRoaring/blob/de7ea45dd1063ad510be24781385f7471d1f7a7d/src/containers/array.c#L519

For `roaring`:
* `cargo fmt -- --check` ✅
* `cargo clippy --all-targets -- -D warnings` ✅
* `cargo test` ✅


Kerollmops (approved):
Nice addition, thank you! Easy enough and nice perf win 👍

